PR #6, opened 2019-08-07
Add scaffolding for tests

jtpio:
Add initial scaffolding for tests.

```bash
jlpm
jlpm run test
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,10 +31,11 @@
   },
   "scripts": {
     "build": "tsc -b",
-    "clean": "rimraf lib",
+    "clean": "rimraf lib && rimraf tsconfig.tsbuildinfo && rimraf tests/build",
     "docs": "typedoc --options tdoptions.json --theme ../../typedoc-theme src",
     "prepublishOnly": "npm run build",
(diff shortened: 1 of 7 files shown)


afshin:
For now, let's not check in `yarn.lock` files. Maybe we should add it to the `.gitignore`?

jtpio:
Sure, we can do that in a separate PR to remove the existing one.

afshin:
Is the `tslint.json` applying itself to the test files?

jtpio:
Then the double quotes should become single quotes, which doesn't seem to be the case.

afshin:
I think `tslint` is passive, so your editor should be complaining?

`jlpm run prettier` is active and will modify your files, though.

jtpio:
VS Code needs the TSLint extension to start complaining.

Otherwise we make prettier run automatically as a git commit hook?

afshin:
Yeah, it's probably a good idea to add it to commit hook.

I'm running VScode with that plugin, but you're right, it should not be required.

jtpio:
Opened #8 for prettier related discussions.

PR #9, opened 2019-08-08
Debug session interface

jtpio:
Provide a simple `DebugSession` interface that can be used to send debug messages to the kernel.

### TODO

- [x] Add `DebugSession`
- [x] Use `xpython` as `kernelPreference`
- [x] Implement `start` method
- [x] Implement `stop` method

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,7 @@ junit.xml
 *.tsbuildinfo
 
 yarn.lock
+
+# xeus-python debug logs
+xpython_debug_logs
+xeus.log
--- a/src/session.ts
+++ b/src/session.ts
@@ -85,23 +85,23 @@ export class DebugSession implements IDebugger.ISession {
    * @param command debug command.
    * @param args arguments for the debug command.
    */
-  async sendRequest<K extends keyof IDebugger.ISession.IRequest>(
+  async sendRequest<K extends keyof IDebugger.ISession.Request>(
     command: K,
-    args: IDebugger.ISession.IRequest[K]
-  ): Promise<IDebugger.ISession.IResponse[K]> {
+    args: IDebugger.ISession.Request[K]
+  ): Promise<IDebugger.ISession.Response[K]> {
     const message = await this._sendDebugMessage({
       type: 'request',
       seq: this._seq++,
       command,
       arguments: args
     });
-    return message.content as IDebugger.ISession.IResponse[K];
+    return message.content as IDebugger.ISession.Response[K];
   }
 
   /**
    * Signal emitted for debug event messages.
    */
-  get eventMessage(): ISignal<DebugSession, IDebugger.ISession.IEvent> {
+  get eventMessage(): ISignal<DebugSession, IDebugger.ISession.Event> {
     return this._eventMessage;
   }
 
@@ -116,7 +116,7 @@ export class DebugSession implements IDebugger.ISession {
     if (msgType !== 'debug_event') {
       return;
     }
-    const event = message.content as IDebugger.ISession.IEvent;
+    const event = message.content as IDebugger.ISession.Event;
     this._eventMessage.emit(event);
   }
 
@@ -143,7 +143,7 @@ export class DebugSession implements IDebugger.ISession {
   private _disposed = new Signal<this, void>(this);
   private _isDisposed: boolean = false;
 
-  private _eventMessage = new Signal<DebugSession, IDebugger.ISession.IEvent>(
+  private _eventMessage = new Signal<DebugSession, IDebugger.ISession.Event>(
     this
   );
 
--- a/src/session.ts
+++ b/src/session.ts
@@ -25,15 +25,25 @@ export class DebugSession implements IDebugger.ISession {
   }
 
   /**
+   * The client session to connect to a debugger.
+   */
+  client: IClientSession;
+
+  /**
+   * The code editors in a debugger session.
+   */
+  editors: CodeEditor.IEditor[];
+
+  /**
    * Dispose the debug session.
    */
   dispose(): void {
     if (this.isDisposed) {
       return;
     }
-    this.client.iopubMessage.disconnect(this._handleEvent, this);
     this._isDisposed = true;
     this._disposed.emit();
+    Signal.clearData(this);
   }
 
   /**
@@ -53,7 +63,7 @@ export class DebugSession implements IDebugger.ISession {
   /**
    * Start a new debug session
    */
-  public async start(): Promise<void> {
+  async start(): Promise<void> {
     await this.sendRequest('initialize', {
       clientID: 'jupyterlab',
       clientName: 'JupyterLab',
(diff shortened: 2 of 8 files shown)


jtpio:
Spent some time trying to use the reference `vscode-debugadapter-node` package "off the shelf", by directly extending from a [`DebugSession`](https://github.com/microsoft/vscode-debugadapter-node/blob/98326c7b9dd26715b2c7050598cbf14ac3cac43c/adapter/src/debugSession.ts#L282).

Although it could be something interesting to do in the long run, it requires dealing with the `node` specific parts in some way (patching or overriding methods).

For now we might be better off using the `DebugProtocol` from `vscode-debugprotocol` and provide a thin layer to send debug messages and listen to debug events (using JupyterLab / Phosphor primitives such such as `Signal`)

afshin:
I think this approach is better in the long-term in any case. Adding our own filter to the upstream means we always need to **opt in** to new features introduced by `vscode-debugprotocol`, which is more flexible over time.

src/session.ts
```diff
@@ -0,0 +1,248 @@
...
+    /**
+     * The client session used by the debug session.
+     */
+    client: IClientSession;
+  }
+
+  /**
+   * Interface for all the debug requests types.
+   */
+  export interface IDebugRequestTypes {
```
afshin: How do you feel about changing this from `DebugSession.IDebugRequestTypes` to `IDebugger.Requests`?

So that you could use it in this way:
```typescript
export const attach = async (
  args: IDebugger.Requests['attach']
): Promise<void> => Promise.resolve(undefined);
```

jtpio:
> Adding our own filter to the upstream means we always need to **opt in** to new features introduced by `vscode-debugprotocol`, which is more flexible over time.

Exactly. And this lets us add one feature at a time.

src/session.ts
afshin: From @jtpio on Gitter:
> don't we want Requests and Responses to be for a session rather than a debugger?

So you are proposing this?

```typescript
const attach = async (
  args: IDebugger.ISession.Requests['attach']
): Promise<IDebugger.ISession.Responses['attach']> => {
  return { body: null } as IDebugger.ISession.Responses['attach'];
};
```

jtpio: Yes that would be the idea

jtpio: I admit it's a bit "nested" though

src/session.ts
```diff
@@ -0,0 +1,260 @@
...
+      command,
+      arguments: args
+    });
+    return message.content as DebugSession.IDebugResponseTypes[K];
+  }
+
+  /**
+   * Signal emitted for debug event messages.
+   */
+  get eventMessage(): ISignal<DebugSession, DebugProtocol.Event> {
```
jtpio: We can consider aliasing `DebugProtocol.Event` to `IDebugger.ISession.Event` here (generic event)

afshin: 👍

src/session.ts
afshin: Yeah, I agree, but I think it does make sense to do it this way even if it is deeper in the namespace.

jtpio: This actually becomes `IDebugger.ISession.IResponses` (leading "I" to follow the tslint rule)

jtpio:
Instructions have been added to the [README](https://github.com/jupyterlab/debugger/blob/aaed345c5bee741f31295db1ad01abcf003e1863/README.md#development) for the local dev setup as well as running the tests.

There is still more details to iterate on, such as whether the `DebugSession` should hold any state, and making use of the code editors. But these can be addressed in follow-up PRs.

jtpio:
To use the `DebugSession` and send debug messages:

```typescript
// client is an IClientSession
let debugSession = new DebugSession({ client });
await debugSession.start();
// ...
// send other debug messages, for example:
const reply = await debugSession.sendRequest('continue', { threadId: 1 });
// reply is a IDebugger.ISession.IResponse['continue']
// ...
await debugSession.stop();
```

(few more examples in the tests)

src/session.ts
afshin: Instead of `IResponses` I was considering `type Responses` and `type Requests`. What are your thoughts?

.gitignore
```diff
@@ -28,3 +28,7 @@ junit.xml
 *.tsbuildinfo
 
 yarn.lock
+
+# xeus-python debug logs
```
afshin: Is this intentional?

src/session.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  /**
+   * Dispose the debug session.
+   */
+  dispose(): void {
+    if (this.isDisposed) {
+      return;
+    }
+    this.client.iopubMessage.disconnect(this._handleEvent, this);
```
afshin: I would propose:
```typescript
dispose(): void {
  if (this.isDisposed) {
    return;
  }
  this._isDisposed = true;
  this._disposed.emit();
  Signal.clearData(this);
}
```

src/session.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Whether the debug session is disposed.
+   */
+  get isDisposed(): boolean {
+    return this._isDisposed;
+  }
+
+  /**
+   * Start a new debug session
+   */
+  public async start(): Promise<void> {
```
afshin: `public` is superfluous

src/session.ts
```diff
@@ -0,0 +1,166 @@
...
+      locale: 'en-us'
+    });
+
+    await this.sendRequest('attach', {});
+  }
+
+  /**
+   * Stop the running debug session.
+   */
+  public async stop(): Promise<void> {
```
afshin: Same here, re: `public`

src/session.ts
```diff
@@ -0,0 +1,166 @@
...
+    const kernel = this.client.kernel;
+    const future = kernel.requestDebug(msg);
+    future.onReply = (msg: KernelMessage.IDebugReplyMsg) => {
+      return reply.resolve(msg);
+    };
+    await future.done;
+    return reply.promise;
+  }
+
+  client: IClientSession;
```
afshin: To conform with JupyterLab conventions, I'd suggest moving these up above the public methods, after the constructor.

src/session.ts
```diff
@@ -0,0 +1,166 @@
...
+  editors: CodeEditor.IEditor[];
+
+  private _disposed = new Signal<this, void>(this);
```
afshin: And also grouping all the private instance variables together (and typically they don't have docstrings)

src/session.ts
jtpio: yes sure, that works.

.gitignore
jtpio: Yes. `xeus-python` needs this folder (for now), and the logs shouldn't be checked in.

src/session.ts
jtpio: Thanks for the tip :+1:

.gitignore
afshin: But isn't that a comment?

jtpio: To give a hint for why we have these log files

jtpio:
Thanks @afshin for looking at it.

I think most of the points have been addressed now.

PR #11, opened 2019-08-09
Debugger-ui

KsavinN:
I'm added a simple layout and model of variables.

Data in variables are mock in VariablesWidget.

I need to create a search component for variables after that I will focus on service.


src/variables/utils/variableTableDescription.ts
```diff
@@ -0,0 +1,114 @@
+import { h, VirtualDOM, VirtualElement } from '@phosphor/virtualdom';
```
saulshanabrook: Are we phasing out the use of `@phosphor/virtualdom` and instead opting to use React?

KsavinN: It depends. Actually, I just start creating a simple layout. Do you think we should stick in use  @phosphor/virtualdom?

afshin: Ah yes, if you're doing this with a virtual DOM then you may as well use React in a `.tsx` file! If you want to make Phosphor `Widget`, `Panel` etc. instances, then that's cool, too. But a future version of Phosphor won't have `@phosphor/virtualdom`.

style/index.css
```diff
@@ -3,8 +3,99 @@
...
+
+
+
+
+.jp-DebuggerSidebar  div:not(:first-child) .debugger-variables__header{
+  border-bottom: solid var(--jp-border-width) var(--jp-border-color1);
+  border-top: solid var(--jp-border-width) var(--jp-border-color1);
+}
+
+.debugger-variables__header {
```
afshin: For the CSS naming conventions, I would propose giving *each* class its own top level name. If they are semantically children of something (for example the variables widget will always live inside a `DebuggerSidebar` then it should look something like this: `.jp-DebuggerSidebar-variables`.

afshin: Or even `.jp-DebuggerVariables` and that way the sub-item could be like `jp-DebuggerVariables-header` etc.

src/variables/utils/variableTableDescription.ts
KsavinN: Hmmm... I didnt know about that. I will fix this.

style/index.css
KsavinN: Ok, cool. I will refactor all css class in index.css

src/variables/utils/variableTableDescription.ts
saulshanabrook: @KsavinN we put together [a guide on using React in JupyterLab](https://jupyterlab.readthedocs.io/en/stable/developer/virtualdom.html). Feel free to ping me if you have any Qs about it?

afshin: Thanks, Saul. I didn't know we had this.

KsavinN:
I replace @phosphor/virtualdom and put new React components.

I steel work with variables box -> the data wich you can see is the Mock set in VariablesWidget.
We can now search all variables by name.

PR #13, opened 2019-08-16
Add tests for stackTrace, scopes, variables

jtpio:
Add several tests for `stackTrace`, `scopes` and `variablesReference`.

They can also serve as a reference for the protocol messages.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -5,7 +5,7 @@ import { IClientSession } from '@jupyterlab/apputils';
 
 import { CodeEditor } from '@jupyterlab/codeeditor';
 
-import { KernelMessage } from '@jupyterlab/services';
+import { KernelMessage, Kernel } from '@jupyterlab/services';
 
 import { PromiseDelegate } from '@phosphor/coreutils';
 
@@ -40,6 +40,9 @@ export class DebugSession implements IDebugger.ISession {
   dispose(): void {
     if (this.isDisposed) {
       return;
+    }
+    if (this._executeFuture) {
+      this._executeFuture.dispose();
     }
     this._isDisposed = true;
     this._disposed.emit();
@@ -88,6 +91,13 @@ export class DebugSession implements IDebugger.ISession {
       restart: false,
       terminateDebuggee: true
     });
+  }
+
+  /**
+   * Request code execution.
+   */
+  async execute(code: string): Promise<void> {
+    void this._sendExecuteMessage({ code });
   }
 
   /**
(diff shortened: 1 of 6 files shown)


jtpio:
This change also adds an `execute` method to the `DebugSession`, used to send a `request_execute` message to the kernel and start debugging the code.

jtpio:
This could be also rebased on https://github.com/jupyterlab/debugger/pull/15 if we get it in first.

src/session.ts
```diff
@@ -90,6 +93,13 @@ export class DebugSession implements IDebugger.ISession {
     });
   }
 
+  /**
+   * Request code execution.
+   */
+  async execute(code: string): Promise<void> {
+    void this._sendExecuteMessage({ code });
```
afshin: Should this be `await` instead of `void`?

jtpio: Right, good catch!

src/session.ts
```diff
@@ -147,8 +157,27 @@ export class DebugSession implements IDebugger.ISession {
...
+  private async _sendExecuteMessage(
+    msg: KernelMessage.IExecuteRequestMsg['content']
+  ): Promise<void> {
+    const kernel = this.client.kernel;
+    if (this._executeFuture) {
+      this._executeFuture.dispose();
+    }
+    this._executeFuture = kernel.requestExecute(msg);
+    this._executeFuture.onReply = (msg: KernelMessage.IExecuteReplyMsg) => {};
+  }
```
afshin: I am confused about the promise this method is returning, in light of my last comment.

What is this promise guaranteeing when it resolves?

jtpio: This should probably be handled in a better way indeed or simplified.

The idea is that an `execute_request` is sent to the kernel to start the debugging session. But the reply message will be received only when all threads have continued.

Also I would say the reply message (`execute_reply`) can be ignored for now.

jtpio: Actually the `execute_request` should be triggered by a user (ctrl-enter a cell), so the `DebugSession` doesn't need this `execute` method.

In the tests we can trigger an `execute_request` by using the kernel directly:

`client.kernel.requestExecute({ code })`

afshin:
Hm, the tests hang for me:

Screen Shot 2019-08-22 at 11 54 19 AM

jtpio:
Looks like it cannot find xeus-python and start it.

jtpio:
Maybe we should make the tests install the spec for `xpython` to the `data_dir` used for the tests instead: https://github.com/jupyterlab/jupyterlab/blob/f1bc02ced61881df94c49929837c49c022f5b115/jupyterlab/tests/test_jupyterlab.py#L126

afshin:
That sounds like a good idea. Hm, I have the kernel when I run lab ...

jtpio:
The last commit installs the kernel spec for xeus-python automatically when running the tests.

Opened https://github.com/jupyterlab/jupyterlab/pull/7089 to make it easier to do the same in other third-party extensions.

jtpio:
@afshin @KsavinN what do you think about getting this one in?

Then we can take care of https://github.com/jupyterlab/debugger/issues/32 and start integrating the debug messages with the UI.

PR #14, opened 2019-08-16
Update some conventions, add license, clean up.

afshin:
Updating the first pass at the UI that @KsavinN  has been working on, changed a few things that are conventional:

* getters/setters should always come *after* `constructor`
* no underscore variables, just camelCase
* add license info header at the top of each code file
* typos
* The type `any` should be a *last resort*.

This is still a work in progress.


src/variables/variable.ts
```diff
@@ -1,29 +1,29 @@
-
-
+// Copyright (c) Jupyter Development Team.
+// Distributed under the terms of the Modified BSD License.
 
 export interface IVariable {
```
jtpio: Maybe we can consider making this interface based on [DebugProtocol.Variable](https://microsoft.github.io/debug-adapter-protocol/specification#Types_Variable)?

KsavinN: I check   DebugProtocol.Variable and I think we can extend this interface or just replace and use DebugProtocol.Variable.

afshin:
There are more things that need to be changed, but I suggest we merge this so we can split it up into smaller tasks:

* The toolbars need to have icons (that either exist in JupyterLab core or that we add like the example one I added here)
* We should switch to using the debug protocol variable as suggested above.
* We should consolidate the components that are simple children (a table/head/body should not be in separate files)
* We should *probably* use the PhosphorJS menu and not a custom one.

afshin:
* The CSS classes need to be normalized.

PR #15, opened 2019-08-19
Add configuration for Azure Pipelines

### LINKED ISSUE #12: Add CI
Since #9 is going to add some tests, we can start thinking about what we want to use to run them automatically.

The 2 main options at the moment seem to be:
- Azure pipelines: already used for the main JupyterLab repo
- [GitHub actions](https://github.com/features/actions): now available in beta, looking promising (and also arguably easier to configure?)

jtpio:
Fixes #12.

Since we're still waiting to get access to GitHub actions, we can instead fall back to azure pipelines in the meantime.

#### Changes:

- Add `jest-junit`
- Use `reporters` config from upstream `testutils`
- Add configuration for azure pipelines. See [Run pipelines with Anaconda environments](https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/anaconda?view=azure-devops&tabs=ubuntu-16-04) for reference.

We need to use conda to be able to install `xeus-python`.
Here is what it looks like with the azure pipelines enabled on my fork: https://github.com/jtpio/debugger/pull/2

We still need to enable it for this repo and that should be enough.
Maybe we can also add other platforms (Windows) later on if that makes sense.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,43 @@
+pool:
+  vmImage: 'ubuntu-16.04'
+
+strategy:
+  matrix:
+    Tests:
+      python.version: '3.7'
+      testResultsFiles: 'tests/**/junit.xml'
+
+steps:
+- bash: echo "##vso[task.prependpath]$CONDA/bin"
+  displayName: Add conda to PATH
+
+- bash: conda create --yes --quiet --name jupyterlab-debugger
+  displayName: Create a new conda environment
+
+- bash: |
+    source activate jupyterlab-debugger
+    conda install --yes --quiet -c conda-forge nodejs xeus-python ptvsd python=$PYTHON_VERSION
+    python -m pip install -U --pre jupyterlab
+  displayName: Install dependencies
+
+- bash: |
+    source activate jupyterlab-debugger
+    jlpm
+    jlpm run build
+  displayName: Build jupyterlab-debugger
+
+- bash: |
+    source activate jupyterlab-debugger
+    export JUPYTER_PATH=${CONDA_PREFIX}/share/jupyter
+    export XEUS_LOG=1
+    jlpm run test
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,43 @@
...
+- bash: |
+    source activate jupyterlab-debugger
+    jlpm
+    jlpm run build
+  displayName: Build jupyterlab-debugger
+
+- bash: |
+    source activate jupyterlab-debugger
+    export JUPYTER_PATH=${CONDA_PREFIX}/share/jupyter
+    export XEUS_LOG=1
```
jtpio: This generates a bit of noise, but is still useful for now.
We can consider disabling the kernel logs later on.

jtpio:
@afshin if you agree with this, it should be possible to enable the Azure Pipelines for this repo from: https://dev.azure.com/jupyterlab/

I would still be happy to switch over to GitHub actions later on if / when we get access to it.

afshin:
@jtpio I am not sure how to enable this. I logged into Azure and could not figure it out.

cc: @saulshanabrook Do you know what to do, here?

saulshanabrook:
@afshin I went to https://dev.azure.com/jupyterlab/jupyterlab/_build?definitionId=1 and click add new pipeline, then found this repo

It seems to be set up now? https://dev.azure.com/jupyterlab/jupyterlab/_build/results?buildId=3002

blink1073:
:rocket: :boom: :fire:

afshin:
Oh sweet! When I went to add a pipeline it should me all the orgs I'm a member of and I was confused. Is that correct? Was I supposed to pick `jupyterlab` and then `debugger`? Thanks, Saul!

blink1073:
Yeah, the UX wasn't super intuitive

jtpio:
Thanks!

PR #29, opened 2019-08-27
Callstack and Breakpoint Development 

KsavinN:
- Add model for Callstack and Breakpoint
- Add body reactWidget for both 
- Add mock data


src/breakpoints/index.ts
```diff
@@ -52,13 +67,76 @@ class BreakpointsHeader extends Widget {
...
+  export class IModel implements IModel {
+    constructor(model: IBreakpoint[]) {
+      this._state = model;
+    }
+
+    get breakpointsChanged(): ISignal<this, IBreakpoint[]> {
+      return this._breakpointsChanged;
+    }
+
+    get activesChange(): ISignal<this, boolean> {
```
jtpio: To follow the same naming convention as for the other signals:

```suggestion
    get activeChanged(): ISignal<this, boolean> {
```

src/breakpoints/index.ts
```diff
@@ -52,13 +67,76 @@ class BreakpointsHeader extends Widget {
 }
 
 export namespace Breakpoints {
+  export interface IBreakpoint extends DebugProtocol.Breakpoint {}
```
jtpio: Maybe this could be a type alias for now if we don't plan to extend the `DebugProtocol.Breakpoint` interface?

```typescript
export type IBreakpoint = DebugProtocol.Breakpoint;
```

KsavinN: Good question. Hmm... It depends on how actually debugger will be work. If we don't wanna implement own UI functionality for breakpoints it's ok.  I will think about this.

src/breakpoints/index.ts
```diff
@@ -52,13 +67,76 @@ class BreakpointsHeader extends Widget {
...
+
+    get activeChanged(): ISignal<this, boolean> {
+      return this._activeChange;
+    }
+
+    get breakpoints(): IBreakpoint[] {
+      return this._state;
+    }
+
+    get isActive(): boolean {
```
jtpio: We will probably want to have control on the active state at the breakpoint level rather than all the breakpoints at once (can be done in follow-up PRs)

src/breakpoints/index.ts
```diff
@@ -22,16 +26,31 @@ export class Breakpoints extends Panel {
       'deactivate',
       new ToolbarButton({
         iconClassName: 'jp-DebuggerDeactivateIcon',
+        tooltip: `${this.isAllAcitve ? 'Deactivate' : 'Activate'} Breakpoints`,
         onClick: () => {
-          console.log('`deactivate` was clicked');
+          this.isAllAcitve = !this.isAllAcitve;
```
jtpio: Little typo spotted:

```suggestion
          this.isAllActive = !this.isAllActive;
```

jtpio:
Thanks @KsavinN, this is looking really good.

I left a quick suggestion, then we should be good to go and continue iterating on it in follow-up PRs.

PR #33, opened 2019-08-29
Expend Variable Model by scopes.

KsavinN:


src/variables/index.ts
```diff
@@ -98,67 +120,81 @@ export namespace Variables {
...
-      description: 'def test2(): return 0'
-    },
-    {
-      name: 'test 3',
-      value: 'function()',
-      type: 'function',
-      variablesReference: 0,
-      description: 'def test1(): return 0'
-    },
+  scope: [
```
jtpio: How about using the plural form?

```suggestion
  scopes: [
```

src/variables/index.ts
```diff
@@ -98,67 +120,81 @@ export namespace Variables {
...
+      ]
     },
     {
-      name: 'test 5',
-      value: 'function()',
-      type: 'function',
-      variablesReference: 0,
-      description: 'def test1(): return 0'
+      name: 'global',
+      variables: [] as Variables.IVariable[]
```
jtpio: Should we add some mock variables here too?

src/variables/index.ts
KsavinN: Yeah, it's obvious. My bad

src/variables/index.ts
KsavinN: I can add some mock variables for another scope. I put there only for the need to see changes in table variables after  changed scope.

src/variables/index.ts
jtpio: All good :+1:

src/variables/index.ts
```diff
@@ -98,67 +120,97 @@ export namespace Variables {
...
     {
-      name: 'test 5',
-      value: 'function()',
-      type: 'function',
-      variablesReference: 0,
-      description: 'def test1(): return 0'
+      name: 'global',
+      variables: [
+        {
+          name: 'exmapleGlobal',
```
jtpio:
```suggestion
          name: 'exampleGlobal',
```

jtpio: A few similar ones below too.

src/variables/body/index.ts
```diff
@@ -40,7 +40,9 @@ class Description extends Widget {
 
     this.model.currentChanged.connect(
       (model: Variables.IModel, variable: Variables.IVariable) => {
-        this.node.innerHTML = this.renderDescription(this.model.current);
+        this.node.innerHTML = this.renderDescription(
```
afshin: Updating the DOM of this widget should happen in an override of `protected onUpdateRequest` and you should only need to call `this.update();`.

For an example, consider this one: https://github.com/jupyterlab/jupyterlab_apod/blob/1.0/src/index.ts#L58-L85

Although in our case, we don't need `onUpdateRequest` to be `async`. It can be simply `protected`.

src/variables/body/search.tsx
```diff
@@ -62,15 +62,18 @@ class SearchInput extends ReactWidget {
 }
 
 class ScopeSearch extends ReactWidget {
-  constructor() {
+  constructor(model: Variables.IModel) {
     super();
+    this.model = model;
     this.node.style.overflow = 'visible';
```
afshin: We can probably move these styles into the style sheet.

PR #36, opened 2019-09-02
Connect Breakpoints widget with NoteBook

KsavinN:
breakpoints


src/breakpoints/index.ts
```diff
@@ -43,17 +47,153 @@ export class Breakpoints extends Panel {
...
+    }
+  }
+
+  protected getCell(): CodeCell {
+    return this.notebook.activeCell as CodeCell;
+  }
+
+  protected removeAllGutterBreakpoints(cell: CodeCell) {
+    const editor = cell.editor as CodeMirrorEditor;
+    editor.editor.getDoc().eachLine(line => {
```
jtpio: Thinking out loud: I was wondering whether there was a way to avoid the `editor.editor` pattern and use the editor wrapper instead (probably not since we need to access the `CodeMirror` methods directly)

src/breakpoints/index.ts
```diff
@@ -7,12 +7,16 @@ import { Widget, Panel, PanelLayout } from '@phosphor/widgets';
...
+import { Editor, Doc } from 'codemirror';
+import { CodeCell } from '@jupyterlab/cells';
 
 export class Breakpoints extends Panel {
-  constructor(options: Breakpoints.IOptions = {}) {
+  constructor(options: Breakpoints.IOptions) {
     super();
 
-    this.model = new Breakpoints.IModel(MOCK_BREAKPOINTS);
+    this.model = new Breakpoints.IModel([] as Breakpoints.IBreakpoint[]);
```
jtpio: Is the type casting necessary here?

src/breakpoints/index.ts
```diff
@@ -43,17 +47,153 @@ export class Breakpoints extends Panel {
...
   }
 
   private isAllActive = true;
   readonly body: Widget;
   readonly model: Breakpoints.IModel;
-}
+  notebook: INotebookTracker;
+  previousCell: CodeCell;
+  previousLineCount: number;
+  cellsBreakpoints: any = {};
```
jtpio: This needs a proper type instead of `any`.

src/breakpoints/index.ts
```diff
@@ -43,17 +47,153 @@ export class Breakpoints extends Panel {
...
+    if (this.previousLineCount !== linesNumber) {
+      if (this.previousLineCount > linesNumber) {
+        this.model.changeLines(lineInfo.line, -1);
+      }
+      if (this.previousLineCount < linesNumber) {
+        this.model.changeLines(lineInfo.line, +1);
+      }
+      this.previousLineCount = linesNumber;
+    }
+    // eage case for backspace line 2
```
jtpio: what does this edge case look like?

src/breakpoints/index.ts
```diff
@@ -43,17 +47,153 @@ export class Breakpoints extends Panel {
...
+      return;
+    }
+    const removeGutter = !!info.gutterMarkers;
+
+    const breakpoint: Breakpoints.IBreakpoint = this.model.getBreakpointByLineNumber(
+      lineNumber
+    );
+
+    if (!breakpoint && !removeGutter) {
+      this.model.breakpoint = {
```
jtpio: This looks like the model can only handle 1 breakpoint at a time. Maybe using a more explicit method such as `addBreakpoint` would be more appropriate?

src/breakpoints/index.ts
```diff
@@ -107,9 +247,35 @@ export namespace Breakpoints {
...
+        this.breakpoints = [];
+        this.breakpoints = breakpoints;
       }
     }
 
+    removeBreakpoint(breakpoint: IBreakpoint) {
+      const breakpoints = this.breakpoints.filter(
+        ele => ele.line !== breakpoint.line
+      );
+      this.breakpoints = [];
```
jtpio: Is this line necessary?

src/breakpoints/index.ts
```diff
@@ -107,9 +247,35 @@ export namespace Breakpoints {
...
+      this.breakpoints = breakpoints;
+    }
+
+    changeLines(lineEnter: number, howMany: number) {
+      const breakpoints = this.breakpoints.map(ele => {
+        ele.line = lineEnter <= ele.line ? ele.line + howMany : ele.line;
+        return ele;
+      });
+      this.breakpoints = [];
```
jtpio: Same as above

src/debugger.ts
```diff
@@ -64,6 +70,10 @@ export namespace Debugger {
       return this._isDisposed;
     }
 
+    get notebook() {
```
jtpio: This seems to return a `NotebookTracker` instead of `Notebook` (can lead to some confusion for users of this API)

src/breakpoints/index.ts
KsavinN: If you set a breakpoint in line 1 and after that remove line 2 in editor gutter remove automatically breakpoint from line 1.

src/breakpoints/index.ts
```diff
@@ -43,17 +47,160 @@ export class Breakpoints extends Panel {
...
   }
 
   private isAllActive = true;
   readonly body: Widget;
   readonly model: Breakpoints.IModel;
-}
+  noteTracker: INotebookTracker;
+  previousCell: CodeCell;
+  previousLineCount: number;
+  cellsBreakpoints: { [id: string]: Breakpoints.IBreakpoint[] } = {};
```
afshin: Could we change the name of this to `cellBreakpoints` where `cell` is singular, please?

src/breakpoints/index.ts
```diff
@@ -43,17 +47,160 @@ export class Breakpoints extends Panel {
...
+  }
+
+  protected removeAllGutterBreakpoints(cell: CodeCell) {
+    const editor = cell.editor as CodeMirrorEditor;
+    editor.editor.getDoc().eachLine(line => {
+      editor.editor.setGutterMarker(line, 'breakpoints', null);
+    });
+  }
+
+  removeListner(cell: CodeCell) {
```
afshin: typo: `removeListner` => `removeListener`

src/debugger.ts
```diff
@@ -45,14 +47,16 @@ export class Debugger extends BoxPanel {
 export namespace Debugger {
   export interface IOptions {
     connector?: IDataConnector<ReadonlyJSONValue>;
-
+    noteTracker?: INotebookTracker;
```
afshin: I'm not sure I understand the reasoning here. Why does the `Debugger` need a reference to any trackers? The pattern I had in mind was the extension to have the trackers and as those trackers change for it to modify the debugger's handle to individual text editors.

KsavinN: In actual branch, i working on a split this. I get your idea before, these changes I made after the suggestion of Jeremy. I guess tomorrow or on Wednesday I will merge changes to this branch.

afshin: Ah okay, great; I'll take a look whenever you're ready for that!

src/index.ts
```diff
@@ -139,11 +178,13 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
     const tracker = new WidgetTracker<MainAreaWidget<Debugger>>({
       namespace: 'debugger'
     });
-
+    tracker.widgetUpdated.connect((_, upadete) => {
+      upadete;
```
```suggestion
      update;
```

src/handlers/notebook.ts
```diff
@@ -0,0 +1,74 @@
...
+          activeCell: codeCell,
+          session: this.debuggerSession
+        };
+        this.cellManager = new CellManager(options);
+      }
+    });
+  }
+
+  protected newDebuggerSession(
+    client: IClientSession | Boolean,
```
jtpio: How about using `null` instead of `Boolean`?

src/handlers/notebook.ts
```diff
@@ -0,0 +1,74 @@
...
+import { CodeCell } from '@jupyterlab/cells';
+import { CellManager } from './cell';
+
+export class DebuggerNotebookTracker {
+  constructor(options: DebuggerNotebookTracker.IOptions) {
+    this.breakpointService = options.breakpointService;
+    this.notebookTracker = options.notebookTracker;
+
+    this.notebookTracker.currentChanged.connect(
+      (sender, notePanel: NotebookPanel) => {
```
jtpio: `notebookPanel` sounds more explicit. What do you think?

```suggestion
      (sender, notebookPanel: NotebookPanel) => {
```

src/handlers/console.ts
```diff
@@ -0,0 +1,93 @@
...
+        this.newDebuggerSession(consolePanel, sender);
+      }
+    );
+  }
+
+  consoleTracker: IConsoleTracker;
+  debuggerSession: DebugSession;
+  breakpointService: BreakpointsService;
+  cellManager: CellManager;
+  consoleSession: IClientSession | Boolean;
```
jtpio: Having `consoleSession: IClientSession | null;` would help avoiding the several casts below.

jtpio:
Quickly tested this locally. The "Remove all Breakpoints" button doesn't remove the breakpoints from the cell gutters, but that can be done in a follow-up PR.

src/breakpoints/index.ts
```diff
@@ -75,44 +77,94 @@ export namespace Breakpoints {
...
+        this._breakpoints[index] = breakpoint;
         this._breakpointChanged.emit(breakpoint);
+      } else {
+        this.breakpoints = [...this.breakpoints, breakpoint];
+      }
+    }
+
+    addBreakpoint(session: string, type: string, lineInfo: LineInfo) {
+      const breakpoint: Breakpoints.IBreakpoint = {
+        line: lineInfo.line,
```
jtpio: This seems to return `0` for line number `1`?

KsavinN: yes, good point. I will fix that

src/breakpoints/index.ts
```diff
@@ -75,44 +77,94 @@ export namespace Breakpoints {
...
-    private _breakpointsChanged = new Signal<this, IBreakpoint[]>(this);
+    changeLines(linesInfo: LineInfo[]) {
+      if (!linesInfo && this.breakpoints.length === 0) {
+        return;
+      }
+      if (linesInfo.length === 0) {
+        this.breakpoints = [];
+      } else {
+        const breakpoint = { ...this.breakpoints[0] };
+        var breakpoints: Breakpoints.IBreakpoint[] = [];
```
jtpio: `let` would be better here:

```suggestion
        let breakpoints: Breakpoints.IBreakpoint[] = [];
```

src/index.ts
```diff
@@ -42,14 +52,20 @@ export namespace CommandIDs {
...
-    }
-    console.log(`${consoles.id} has not been implemented.`, debug);
+  requires: [IDebugger, IDebuggerSidebar, IConsoleTracker],
+  activate: (
+    _,
+    debug: IDebugger,
+    sidebar: IDebuggerSidebar,
+    tracker: IConsoleTracker
+  ) => {
+    debug.currentChanged.connect((_, update) => {
```
jtpio: Is this signal actually emitted?

src/index.ts
```diff
@@ -153,7 +220,10 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
         }
 
         const widget = new MainAreaWidget({
-          content: new Debugger({ connector: state, id })
+          content: new Debugger({
```
jtpio: Let's add the widget to the main area?

For now it's only added to the tracker.

jtpio:
Sometimes it seems like the breakpoints can't be set (after running the `debugger:create` command and refreshing the page).

(here the panel was added to the main area with `app.shell.add(widget, 'main')`)

breakpoints-gutter

src/breakpoints/index.ts
jtpio: :+1:

src/index.ts
KsavinN: I think we could do that. But I don't quite know what we can now display in that widget.

jtpio: Not much for now indeed, but it's still good that it shows up somewhere.

KsavinN: Ok. I will change that and try to resolve bug which you found.

afshin:
I propose we merge this and iterate on it so that @jtpio and I can work off this branch.

JohanMabille:
I can reproduce the bug reporter by @jtpio in a deterministic way: I can never set any breakpoint (I have checked that the extension is correctly installed and enabled). Also for a unknown reason, the display of line numbers is disabled by default.

I've tested with [this prototype](https://github.com/QuantStack/jupyterlab-debugger), I can set breakpoints in cells that have not been executed yet only.

KsavinN:
> I can reproduce the bug reported by @jtpio in a deterministic way: I can never set any breakpoint (I have checked that the extension is correctly installed and enabled). Also for an unknown reason, the display of line numbers is disabled by default.
> 
> I've tested with [this prototype](https://github.com/QuantStack/jupyterlab-debugger) (to see if I could quickly port a fix) I can set breakpoints, but only in cells that have not been executed yet.
> 
> NOTE: I'm testing with Google Chrome.

@JohanMabille Thank you for help and opinion.  I will try to reproduce issues. "Also for an unknown reason, the display of line numbers is disabled by default." - I was thought that the default state of display line numbers is disabled.

JohanMabille:
@KsavinN The bug I had was my fault, I didn't start the debugger with the "hidden" command. Now I can se breakpoints and reproduce the issue described by @jtpio

package.json
```diff
@@ -52,6 +52,7 @@
     "@phosphor/coreutils": "^1.3.1",
     "@phosphor/disposable": "^1.2.0",
     "@phosphor/widgets": "^1.8.0",
+    "@types/codemirror": "0.0.76",
```
jtpio: This should go to `devDependencies`

PR #46, opened 2019-10-03
Fixed debugger restore

JohanMabille:
This fixes the problem described by @jtpio [here](https://github.com/jupyterlab/debugger/pull/36#issuecomment-537105099).

The remaining issue is that executing the debugger command creates a new debugger instance while it should reuse the existing one.


jtpio:
I also think we should keep the `UUID` generation for now, even though new debugger `MainAreaWidgets` are created every time we run the command.

We can look at how to have a single panel (and find an existing one debugger main area widget) in follow-up PRs.

PR #47, opened 2019-10-04
Connect to debbuger 

KsavinN:
Work in progress

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -228,9 +228,14 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
 
     app.commands.addCommand(CommandIDs.start, {
       label: 'Start',
+      isEnabled: () => {
+        const debuggerModel = getModel();
+        return (debuggerModel &&
+          debuggerModel.session !== undefined) as boolean;
+      },
       execute: async () => {
         const debuggerModel = getModel();
-        if (debuggerModel) {
+        if (debuggerModel && debuggerModel.session) {
           await debuggerModel.session.start();
           commandStop = palette.addItem({
             command: CommandIDs.stop,
--- a/src/index.ts
+++ b/src/index.ts
@@ -208,19 +208,13 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
       update;
     });
 
-    const commandCreateDebugger = CommandIDs.create;
-
-    const commandStartDebugger = CommandIDs.start;
-
-    const commandStopDebugger = CommandIDs.stop;
-
-    var commandStop: IDisposable;
+    let commandStop: IDisposable;
 
     const getModel = () => {
       return tracker.currentWidget ? tracker.currentWidget.content.model : null;
     };
 
-    app.commands.addCommand(commandStopDebugger, {
+    app.commands.addCommand(CommandIDs.stop, {
       label: 'Stop',
       execute: async () => {
         const debuggerModel = getModel();
@@ -232,21 +226,21 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
       }
     });
 
-    app.commands.addCommand(commandStartDebugger, {
+    app.commands.addCommand(CommandIDs.start, {
       label: 'Start',
       execute: async () => {
         const debuggerModel = getModel();
         if (debuggerModel) {
           await debuggerModel.session.start();
-          commandStop = await palette.addItem({
-            command: commandStopDebugger,
+          commandStop = palette.addItem({
+            command: CommandIDs.stop,
             category: 'Debugger'
           });
         }
       }
     });
 
-    app.commands.addCommand(commandCreateDebugger, {
+    app.commands.addCommand(CommandIDs.create, {
       label: 'Debugger',
       execute: args => {
         const id = (args.id as string) || UUID.uuid4();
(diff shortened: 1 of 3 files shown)


KsavinN:
I added command in pallet to run debugger (like start and stop).

During development I found a bug :

 - Session start sometimes doesn't work after reload for an existing notebook.

jtpio:
Thanks @KsavinN, will have a look at it soon!

src/index.ts
```diff
@@ -212,22 +208,55 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
...
+      }
+    });
+
+    app.commands.addCommand(commandStartDebugger, {
+      label: 'Start',
+      execute: async () => {
+        const debuggerModel = getModel();
+        if (debuggerModel) {
+          await debuggerModel.session.start();
+          commandStop = await palette.addItem({
```
jtpio: Does `palette.addItem` return a promise?

src/index.ts
```diff
@@ -245,12 +274,16 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
       }
     });
 
-    palette.addItem({ command, category: 'Debugger' });
+    palette.addItem({ command: commandCreateDebugger, category: 'Debugger' });
+    palette.addItem({
+      command: commandStartDebugger,
+      category: 'Debugger'
+    });
```
jtpio: Since `palette` is passed as an optional dependency, we should first check the palette is present before adding items to it:

```typescript
if (palette) {
  palette.addItem(...);
}
```

src/session.ts
```diff
@@ -149,6 +149,7 @@ export class DebugSession implements IDebugger.ISession {
     sender: IClientSession,
     message: KernelMessage.IIOPubMessage
   ): void {
+    // console.log({ sender, message });
```
jtpio: To be removed?

src/index.ts
```diff
@@ -212,22 +208,55 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
...
+
+    const commandStopDebugger = CommandIDs.stop;
+
+    var commandStop: IDisposable;
+
+    const getModel = () => {
+      return tracker.currentWidget ? tracker.currentWidget.content.model : null;
+    };
+
+    app.commands.addCommand(commandStopDebugger, {
```
jtpio: How about directly using `CommandIDs.stop` here? (same question for the others commands)

src/index.ts
```diff
@@ -212,22 +208,55 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
...
     });
 
-    const command = CommandIDs.create;
+    const commandCreateDebugger = CommandIDs.create;
+
+    const commandStartDebugger = CommandIDs.start;
+
+    const commandStopDebugger = CommandIDs.stop;
+
+    var commandStop: IDisposable;
```
```suggestion
    let commandStop: IDisposable;
```

jtpio:
>     * Session start sometimes doesn't work after reload for an existing notebook.

What does 'reload' mean in this context?

This is probably because the debugger is already started on the kernel side, and refreshing the page (or reloading the notebook) doesn't send (and shouldn't) a disconnect request.

jtpio:
@KsavinN also, it's possible to enable the kernel logs by setting the `XEUS_LOG=1` environment variable before starting JupyterLab:

`XEUS_LOG=1 jupyter lab --no-browser --watch`

There should be some messages when the debugger is started and stopped.

This is a good way to see what is happening on the kernel side.

KsavinN:
> > ```
> > * Session start sometimes doesn't work after reload for an existing notebook.
> > ```
> 
> What does 'reload' mean in this context?
> 
> This is probably because the debugger is already started on the kernel side, and refreshing the page (or reloading the notebook) doesn't send (and shouldn't) a disconnect request.

After reloading the page.  @jtpio Thanks for the advice I will check what happened by this:

> @KsavinN also, it's possible to enable the kernel logs by setting the `XEUS_LOG=1` environment variable before starting JupyterLab:
> 
> `XEUS_LOG=1 jupyter lab --no-browser --watch`
> 
> There should be some messages when the debugger is started and stopped.
> 
> This is a good way to see what is happening on the kernel side.

jtpio:
The logs look like this:

```
msg_type: debug_request
XEUS: sent message on iopub - kernel_core.99ffbc8efa244f46ac8da35b410e3e44.status
msg_type: status
XEUS-PYTHON: the debugger has started
XEUS: sent message on control - e7eccdc6-8028-4ea3-9573-bef39b189253
```

jtpio:
Ideally the `start` command should be disabled (shown but not enabled) when a debug session is already available on the kernel side.

jtpio:
Thanks @KsavinN, let's merge this one.

One thing to look at next would be to add the debugger widget back to the main area.

jtpio:
So testing this locally was working fine because a workspace with a Debugger object was already available.

But clearing the workspace (`Reset application state`), which removes tracked widgets, also seems to be breaking the start / stop commands.

Something that @JohanMabille is also experiencing locally on master.

@KsavinN let me know if you want to have a look at it, otherwise I'll investigate further tomorrow.

KsavinN:
@jtpio  You can investigate it tomorrow. For now, I focus on  PR of Darian https://github.com/jupyterlab/debugger/pull/48.

jtpio:
ok. this PR needs rebase / merge though.

PR #53, opened 2019-10-14
Start and stop always visible, enabled/disabled

JohanMabille:
- ~set ILayoutRestorer and ICommandPalette arguments as required~
- always displays the start and stop commands of the debugge
- start is enabled when a debugging session is available and the debugger has not started yet, disabled otherwise
- stop is enabled when a debugging session is available and the debugger has started, disabled otherwise.

Remaining issues:
- the '_isStarted' cache variable in the DebuggerSession object should be updated upon event notification
- the current implementation creates a new DebuggerSession each time a notebook gets the focus instead of restoring a previous existing session. This breaks the mechanism for enabling and disabling the start and stop commands. The session should be saved somewhere and restored, but I don't know yet the best way to achieve this.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,8 +32,6 @@ import { DebuggerNotebookHandler } from './handlers/notebook';
 
 import { DebuggerConsoleHandler } from './handlers/console';
 
-import { IDisposable } from '@phosphor/disposable';
-
 import { Kernel } from '@jupyterlab/services';
 
 /**
@@ -224,7 +222,6 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
     });
     const { commands, shell } = app;
     let widget: MainAreaWidget<Debugger>;
-    let commandStop: IDisposable;
 
     const getModel = () => {
       return tracker.currentWidget ? tracker.currentWidget.content.model : null;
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -189,16 +188,16 @@ const sidebar: JupyterFrontEndPlugin<IDebuggerSidebar> = {
  */
 const tracker: JupyterFrontEndPlugin<IDebugger> = {
   id: '@jupyterlab/debugger:tracker',
-  optional: [ILayoutRestorer, IDebuggerSidebar, ICommandPalette],
-  requires: [IStateDB],
+  optional: [IDebuggerSidebar],
+  requires: [IStateDB, ICommandPalette, ILayoutRestorer],
```
jtpio: Any reason to move them to `requires`?

JohanMabille: We want to be able to restore the layout and we want the commands to be visible in the command palette, so I don't see any reason for having them as optional (but maybe there is some that I am not aware of).

jtpio: There could be custom builds of JupyterLab that don't provide these tokens (for example no command palette).

JohanMabille: ok, I'll move them back to optional then.

JohanMabille:
@jtpio This is ready for a review, let's fix the remaining issues in dedicated PRs since it may require substantial changes.

jtpio:
> Let's fix the remaining issues in dedicated PRs since it may require substantial changes.

Thanks. Let's create new issues for that.

src/session.ts
```diff
@@ -87,10 +87,18 @@ export class DebugSession implements IDebugger.ISession {
...
+   */
+  get isStarted(): boolean {
+    return this._isStarted;
+  }
+
   /**
    * Start a new debug session
    */
   async start(): Promise<void> {
+    this._isStarted = true;
```
jtpio: Maybe this should be after the call to `attach`?

jtpio: Sound good, using the debug event is indeed better in the long run :+1:

PR #48, opened 2019-10-04
Iterate on interfaces, tokens, and UI.

afshin:

PR #54, opened 2019-10-14
Iterate on interfaces, tokens, and UI #2

KsavinN:
This PR is based on https://github.com/jupyterlab/debugger/pull/48

Only one task was not done: " The Debugger extends BoxPanel class should have an attribute called editors that is a TabPanel of read-only text editors that allow users to add breakpoints."


jtpio:
Thanks @KsavinN for opening the new PR (which replaces #48)

What happens when you start from a fresh workspace? (the `Reset application state` command)

src/handlers/cell.ts
```diff
@@ -32,6 +34,18 @@ export class CellManager {
...
   private breakpointsModel: Breakpoints.Model;
   private _activeCell: CodeCell;
+  isDisposed: boolean;
+
+  dispose(): void {
+    if (this.isDisposed) {
+      return;
+    }
+    if (this.previousCell) {
+      this.removeListner(this.previousCell);
```
jtpio:
```suggestion
      this.removeListener(this.previousCell);
```

src/index.ts
```diff
@@ -84,25 +111,29 @@ const consoles: JupyterFrontEndPlugin<void> = {
...
-        debugModel.sidebar.breakpoints.model.type = session.type as SessionTypes;
+    let _model: any;
+    labShell.currentChanged.connect((_, update) => {
+      const widget = update.newValue;
+      if (!(widget instanceof FileEditor)) {
+        return;
       }
+      const sessions = app.serviceManager.sessions;
+
+      sessions.findByPath((widget as FileEditor).context.path).then(model => {
```
jtpio: This seems to be a floating promise.

src/index.ts
```diff
@@ -14,27 +14,25 @@ import { WidgetTracker, MainAreaWidget } from '@jupyterlab/apputils';
...
-
 import { DebuggerConsoleHandler } from './handlers/console';
-
-import { DebuggerSidebar } from './sidebar';
-
-import { SessionTypes } from './breakpoints';
-import { DebugSession } from './session';
 import { IDisposable } from '@phosphor/disposable';
+import { Kernel } from '@jupyterlab/services';
+// import { Session } from '@jupyterlab/services';
```
jtpio: This can be removed.

src/tokens.ts
```diff
@@ -16,26 +16,47 @@ import { IObservableDisposable } from '@phosphor/disposable';
...
+   * as a `MainAreaWidget`, otherwise it is a sidebar.
+   */
+  mode: IDebugger.Mode;
+
+  /**
+   * The current debugger session.
+   */
+  session: IDebugger.ISession;
+
+  //tracker for get instance of debugger
```
jtpio: This comment should be formatted like the others.

src/index.ts
```diff
@@ -245,52 +300,109 @@ const tracker: JupyterFrontEndPlugin<IDebugger> = {
...
+        return !!tracker.currentWidget;
+      },
+      execute: () => {
+        const currentMode = tracker.currentWidget.content.model.mode;
+        tracker.currentWidget.content.model.mode =
+          currentMode === 'expanded' ? 'condensed' : 'expanded';
+        let mode = tracker.currentWidget.content.model.mode;
+
+        if (mode === 'condensed') {
+          commands.execute(CommandIDs.mount, { mode });
```
jtpio: These calls to `execute` should be prefixed with `void`

jtpio:
> What happens when you start from a fresh workspace? (the `Reset application state` command)

Never mind that. I was able to get the same behavior that you demoed today after cleaning and rebuilding.

Also for future PRs it could be useful to have a quick screencast to get a quick summary of the changes, if they involve visual elements :)

src/tokens.ts
```diff
@@ -16,26 +16,49 @@ import { IObservableDisposable } from '@phosphor/disposable';
...
+
+  /**
+   * The current debugger session.
+   */
+  session: IDebugger.ISession;
+
+  /**
+   * tracker for get instance of debugger.
+   */
+  tracker: WidgetTracker<MainAreaWidget<Debugger>>;
```
jtpio: Why do we need the tracker here? Shouldn't the interface only expose `mode` and `session`?

KsavinN: For handlers in consoles and notebooks Plugin. they need access to the model of Debugger.
Should we create some different way to share model between plugins?

KsavinN:
> > What happens when you start from a fresh workspace? (the `Reset application state` command)
> 
> Never mind that. I was able to get the same behavior that you demoed today after cleaning and rebuilding.
> 
> Also for future PRs it could be useful to have a quick screencast to get a quick summary of the changes, if they involve visual elements :)

i will have that in mind.

src/tokens.ts
jtpio: Yeah ideally we would require the tracker in the plugin (with `requires`), but at the moment it's not provided anywhere.

jtpio: I see that it was changed in https://github.com/jupyterlab/debugger/pull/48 as well.

JohanMabille:
Why adding stuff already done in dedicated PR (https://github.com/jupyterlab/debugger/pull/53) to this one?

KsavinN:
> Why reimplementing stuff already done in dedicated PR (#53)?

@JohanMabille you have right. I will rollback.

jtpio:
@KsavinN let me know when this is ready for another look, so we can get #53 in right after too.

KsavinN:
> @KsavinN let me know when this is ready for another look, so we can get #53 in right after too.

I think now is ready. For now, I don't have idea where we can implement a provider for a tracker of Debugger.

src/index.ts
```diff
@@ -84,25 +113,29 @@ const consoles: JupyterFrontEndPlugin<void> = {
...
+      const widget = update.newValue;
+      if (!(widget instanceof FileEditor)) {
+        return;
       }
+
+      //  Finding if the file is backed by a kernel or attach it to one.
+
+      const sessions = app.serviceManager.sessions;
+
+      sessions.findByPath(widget.context.path).then(model => {
```
jtpio: Floating promise, would be better to prefix it with `void` or use async / await

KsavinN:
@jtpio  I resolved floating promises and fix not saveing mode of Debugger.

jtpio:
Thanks @KsavinN for the fix. I will do a quick test again and then we can continue development in a different PR (to not make this one too big).

jtpio:
>  For now, I don't have idea where we can implement a provider for a tracker of Debugger.

Let's tackle that separately.

PR #57, opened 2019-10-15
Fix linting

jtpio:
Fix a few linting errors.

We should add a step in the test pipeline that does this check or a commit hook like prettier.


jtpio:
cc @KsavinN @JohanMabille

KsavinN:
> cc @KsavinN @JohanMabille

good idea, it's seems to look good. I think we can merge it.

jtpio:
Good, then let's do that.

Issue #38, opened 2019-09-03
Implement debugger expanded layout


afshin:
The `expanded` mode of the debugger should have the debugger sidebar inside a `SplitPanel` that has a `TabPanel` as its other pane (to hold read-only text editors for stepping into code).

PR #61, opened 2019-10-16
Add editors to the main area widget

### LINKED ISSUE #56: Add editors to the debugger main area widget
To be done after https://github.com/jupyterlab/debugger/pull/54 is merged.

Quoting https://github.com/jupyterlab/debugger/pull/48#issue-324688768: 

> The Debugger extends BoxPanel class should have an attribute called editors that is a TabPanel of read-only text editors that allow users to add breakpoints

jtpio:
Fixes #56.

The typical usage would involve calling the `addEditor` method when receiving a message from kernel that we hit a breakpoint in a specific file. `addEditor` will show the content of the file that we stepping into (or a previous cell that was executed and then deleted).

This PR does not cover adding breakpoints to the read-only editors (can be done separately).

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -15,7 +15,7 @@ import { ISignal, Signal } from '@phosphor/signaling';
 
 import { BoxPanel } from '@phosphor/widgets';
 
-import { CodeEditors } from './editors';
+import { DebuggerEditors } from './editors';
 
 import { DebuggerSidebar } from './sidebar';
 
@@ -33,13 +33,13 @@ export class Debugger extends BoxPanel {
     this.model.sidebar = this.sidebar;
 
     const { editorFactory } = options;
-    this.editors = new CodeEditors({ editorFactory });
+    this.editors = new DebuggerEditors({ editorFactory });
     this.addWidget(this.editors);
 
     this.addClass('jp-Debugger');
   }
 
-  readonly editors: CodeEditors;
+  readonly editors: DebuggerEditors;
   readonly model: Debugger.Model;
   readonly sidebar: DebuggerSidebar;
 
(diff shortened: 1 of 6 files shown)


jtpio:
Quick look at the current status (using fake data):

code-editors

Still needs a bit of work on the interfaces, model and so on.

jtpio:
@KsavinN @JohanMabille 

I would like to limit the scope of this PR to just adding read-only editors to the debugger main area widget (current state).

Let me know what you think.

Adding breakpoints and handling messages from the kernel can be done in different PRs.

Also since the `expanded` mode removes the sidebar, we should add a way to make the breakpoints, callstack and variables available from the main widget area. I'm not sure the [current design document](https://github.com/jupyterlab/debugger/blob/master/DESIGN.md) covers this.

jtpio:
> Also since the `expanded` mode removes the sidebar, we should add a way to make the breakpoints, callstack and variables available from the main widget area. I'm not sure the [current design document](https://github.com/jupyterlab/debugger/blob/master/DESIGN.md) covers this.

This is tracked in https://github.com/jupyterlab/debugger/issues/38.

KsavinN:
> @KsavinN @JohanMabille
> 
> I would like to limit the scope of this PR to just adding read-only editors to the debugger main area widget (current state).
> 
> Let me know what you think.
> 
> Adding breakpoints and handling messages from the kernel can be done in different PRs.
> 
> Also since the `expanded` mode removes the sidebar, we should add a way to make the breakpoints, callstack and variables available from the main widget area. I'm not sure the [current design document](https://github.com/jupyterlab/debugger/blob/master/DESIGN.md) covers this.

Agree. We should try to make the smallest PR.

jtpio:
Yes, will be adding the `sidebar` as mentioned in https://github.com/jupyterlab/debugger/issues/38 right after.

jtpio:
:+1: 

Follow-up PR will add the sidebar to the main area widget:

image

PR #63, opened 2019-10-16
enable and disable commands based on replies

### LINKED ISSUE #59: Use debug events to control the state of the session
Quoting https://github.com/jupyterlab/debugger/pull/53#issue-327671083:

> the '_isStarted' cache variable in the DebuggerSession object should be updated upon event notification

We can make use of the [debug events](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Initialized) instead to set `_isStarted`

JohanMabille:
Fixes #59

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -98,32 +98,41 @@ export class DebugSession implements IDebugger.ISession {
    * Start a new debug session
    */
   async start(): Promise<void> {
-    this._isStarted = true;
-    await this.sendRequest('initialize', {
-      clientID: 'jupyterlab',
-      clientName: 'JupyterLab',
-      adapterID: 'python',
-      pathFormat: 'path',
-      linesStartAt1: true,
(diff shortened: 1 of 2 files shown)


src/session.ts
```diff
@@ -98,32 +98,41 @@ export class DebugSession implements IDebugger.ISession {
...
+        supportsVariablePaging: true,
+        supportsRunInTerminalRequest: true,
+        locale: 'en-us'
+      });
+
+      this._isStarted = true;
+
+      await this.sendRequest('attach', {});
+    } catch (err) {
+      console.log('Error: ', err.message);
```
jtpio: What to you think about using `console.error` instead?

JohanMabille: Indeed that would be better. I'll update the PR

jtpio:
Thanks!

PR #65, opened 2019-10-17
Add .prettierignore file

jtpio:
To ignore some folders when formatting code automatically.


jtpio:
Merging now. Other folders can easily be added later if needed.

Issue #44, opened 2019-09-30
Request variables state


jtpio:
Let's replace the fake variables by real ones read from the kernel.
This means implementing the `scopes` and `variables` requests.

There is an example of such requests in the tests:

https://github.com/jupyterlab/debugger/blob/23ed40d5711cd9b1522a893314f7bafd5b97bb7a/tests/src/session.spec.ts#L188-L202

jtpio:
First implementation done in https://github.com/jupyterlab/debugger/pull/66

Issue #64, opened 2019-10-17
Debug Protocol Overview


jtpio:
Adding a quick overview of the debug protocol, so it's easier to visualize the types of messages sent between the JupyterLab extension and the kernel.

It only shows a subset of the messages for now, but can be extended to also include the calls to `continue`, `stepIn` and `disconnect`.

This also show new messages that are not part of the Debug Adapter Protocol (DAP), such as `dumpCell`.

### Diagram

![debug-protocol](https://user-images.githubusercontent.com/591645/71578708-926c0f00-2af9-11ea-81cf-dab838f3dc76.png)

### TODO

- [x] Add the call to `debug_info_request`
- [x] Add the flow to show the files being debugged (stepIn / breakpoints)

### References

- Dump cell and state restoration: https://github.com/jupyterlab/debugger/issues/52
- Protocol Overview: https://microsoft.github.io/debug-adapter-protocol/overview
- Specification: https://microsoft.github.io/debug-adapter-protocol/specification

### Source

Generated using: https://bramp.github.io/js-sequence-diagrams/

<details><summary>Diagram source</summary>
<pre>
user->JupyterLab: open notebook

JupyterLab->kernel: kernel_info_request

kernel->JupyterLab: kernel_info_reply

JupyterLab->JupyterLab: check 'debugger' key

JupyterLab->user: show toggle button\nif 'debugger'

user->JupyterLab: enable debugging

JupyterLab->kernel: debugInfo request

kernel->JupyterLab: debugInfo response

user->JupyterLab: start debugger

JupyterLab->kernel: initialize request

kernel->JupyterLab: initialize response

JupyterLab->kernel: attach request

kernel->JupyterLab: attach response

Note right of kernel: debugger started

user->JupyterLab: add breakpoints\n(click on gutters)

JupyterLab->kernel: dumpCell request

kernel->JupyterLab: dumpCell response

JupyterLab->kernel: setBreakpoints request

kernel->JupyterLab: breakpoints response

JupyterLab->kernel: configurationDone request

kernel->JupyterLab: configurationDone response

user->JupyterLab: execute cell\n(Ctrl-Enter)

JupyterLab->kernel: requestExecute

kernel->kernel: execute code

kernel->kernel: hit breakpoint

kernel-->JupyterLab: stopped event

JupyterLab->kernel: stackTrace request

kernel->JupyterLab: stackTrace response

JupyterLab->user: show current line

JupyterLab->kernel: scopes request

[...]

PR #66, opened 2019-10-17
Create Service for control signals between session and UI

KsavinN:
chrome-capture (2)

For now, we can get scopes,variables by command "Launch".  
The proper work is described in #64 .


src/variables/body/search.tsx
```diff
@@ -93,16 +96,25 @@ const useOutsideClick = (ref: any, callback: any) => {
 
 const ScopeMenuComponent = ({ model }: { model: Variables.IModel }) => {
   const [toggleState, setToggle] = useState(false);
+  const [scopes, setScopes] = useState(model.scopes);
   const [scope, setScope] = useState(model.currentScope);
```
jtpio: Do we need both `scope` and `scopes` in the state?

KsavinN: Scopes can be changed. For now, we get only Local scope or an empty array.

src/variables/body/table.tsx
```diff
@@ -103,12 +113,12 @@ const useTbody = (items: Array<any>, defaultState: any) => {
...
               <td style={{ paddingLeft: `${12}px`, width: `${25}%` }}>
                 <span
-                  className={`jp-Icon jp-Icon-16 ${setClassIcon(item.value)}`}
+                  className={`jp-Icon jp-Icon-16 ${setClassIcon(item.type)}`}
                 ></span>
                 {item.name}
               </td>
               <td style={{ paddingLeft: `${12}px`, width: `${75}%` }}>
-                {item.value}
+                {item.type}
```
jtpio: Shouldn't we keep `item.value` here?

src/variables/index.ts
```diff
@@ -94,19 +99,28 @@ export namespace Variables {
         return;
       }
       this._filterState = value;
-      this._variablesChanged.emit(this._filterVariables());
+      if (!!this._currentScope) {
```
jtpio: Do we need `!!` here?

src/variables/body/table.tsx
KsavinN: I check the response and type there is much better. The value probably we should insert in description

KsavinN:
Resolved issue #73

src/variables/body/table.tsx
jtpio: But `item.type` returns something like `int`, that is shown in the "Value" column.

image

src/service.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export class DebugService {
+  constructor(session: DebugSession | null, debuggerModel: Debugger.Model) {
+    this.session = session;
+    this._model = debuggerModel;
+  }
+
+  private _session: DebugSession;
+  private _model: Debugger.Model;
+  frames: any[];
```
jtpio: Should this be `DebugProtocol.StackFrame[]`?

KsavinN: It's different its have additional value with converted scope for UI. I will add some interface for that

src/service.ts
```diff
@@ -0,0 +1,147 @@
...
+      this.frames.push({
+        id: frame.id,
+        scopes: values
+      });
+      if (index === 0) {
+        this._model.sidebar.variables.model.scopes = values;
+      }
+    });
+
+    if (!!stackFrames) {
```
jtpio: Is `!!` necessary here?

src/index.ts
```diff
@@ -357,11 +373,20 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
...
     if (palette) {
-      palette.addItem({ command: CommandIDs.changeMode, category: 'Debugger' });
-      palette.addItem({ command: CommandIDs.create, category: 'Debugger' });
-      palette.addItem({ command: CommandIDs.start, category: 'Debugger' });
-      palette.addItem({ command: CommandIDs.stop, category: 'Debugger' });
+      palette.addItem({
+        command: CommandIDs.changeMode,
+        category: categoryLabel
+      });
+      palette.addItem({ command: CommandIDs.create, category: categoryLabel });
```
jtpio: This could be become `palette.addItem({ command: CommandIDs.create, category });` by naming the variable `category` instead of `categoryLabel`

jtpio:
@KsavinN is there anything else you wanted to add before we could wrap it up?

We've got initial structure for viewing the state of the variables which is good. We could iterate on the rest in a separate change.

jtpio:
This will fix https://github.com/jupyterlab/debugger/issues/44

KsavinN:
> @KsavinN is there anything else you wanted to add before we could wrap it up?
> 
> We've got the initial structure for viewing the state of the variables which is good. We could iterate on the rest in a separate change.

@jtpio  I will just change code by your suggestion and end that PR. After that, I want focus already on implement react-inspector

KsavinN:
@jtpio  Are you think we can merge that?

jtpio:
I will do a quick check locally first.

src/variables/body/search.tsx
```diff
@@ -92,45 +95,62 @@ const useOutsideClick = (ref: any, callback: any) => {
...
+  //   <ul>
+  //     {!!scopes
+  //       ? scopes.map(scope => (
+  //           <li key={scope.name} onClick={e => changeScope(scope)}>
+  //             {scope.name}
+  //           </li>
+  //         ))
+  //       : null}
+  //   </ul>
+  // );
```
jtpio: Any reason for keeping these comments around?

KsavinN: I forgot about this. Sry

jtpio: :+1:

src/service.ts
```diff
@@ -0,0 +1,147 @@
...
+        name: scope.name,
+        variables: variables.map(variable => {
+          return { ...variable, description: '' };
+        })
+      };
+    });
+  };
+}
+
+export type FrameStored = {
```
jtpio: How about `Frame` only?

jtpio:
Merging as is, and opened these issues to follow-up:

- https://github.com/jupyterlab/debugger/issues/84
- https://github.com/jupyterlab/debugger/issues/85
- https://github.com/jupyterlab/debugger/issues/86
- https://github.com/jupyterlab/debugger/issues/87

PR #68, opened 2019-10-18
Cleanup signals on unmount in BreakpointComponent

jtpio:
Fixes the following issue with unmounted `BreakpointComponent` (React error on unmounted component):

react-breakpoints-component-unmount


jtpio:
Also noticed an issue with the breakpoints listeners, opened a separate issue for that: https://github.com/jupyterlab/debugger/issues/69

PR #72, opened 2019-10-18
Add tslint config for sorted imports

### LINKED ISSUE #31: Automated sorting of imports
One thing that the current tslint / prettier configuration is not covering is the automatic sorting of imports.

It doesn't seem to be enforced in the jupyterlab repo either.

Maybe we can consider having that triggered on `git commit` too.

jtpio:
Fixes #31 

This should automatically sort imports on git commits for the files that have been changed.

We can also consider formatting all the existing files in one PR by running `jlpm run tslint` once.

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -109,6 +109,13 @@
       "check-operator",
       "check-separator",
       "check-type"
+    ],
+    "ordered-imports": [
+      true,
+      {
+        "import-sources-order": "lowercase-last",
+        "named-imports-order": "lowercase-first"
+      }
     ]
   },
   "linterOptions": {


jtpio:
@afshin I realized afterwards that TSLint is going to be deprecated soon (to transition to ESLint): https://medium.com/palantir/tslint-in-2019-1a144c2317a9

Still, it might be a useful thing to have for the time being.

afshin:
I think it's useful to have it for as long as the core JupyterLab packages do, since this extension should end up there.

jtpio:
> I think it's useful to have it for as long as the core JupyterLab packages do, since this extension should end up there.

:+1: 

Although this option (`ordered-imports`) is not part of the tslint / esling config for core JupyterLab. So there might be some slight differences.

PR #78, opened 2019-10-21
Refactored plugins to avoid code duplication

JohanMabille:


jtpio:
Thanks @JohanMabille.

The code for the two plugins is indeed very similar. I'm just wondering if this is going to be the case later on.

src/handlers/console.ts
```diff
@@ -14,9 +14,9 @@ import { IDisposable } from '@phosphor/disposable';
 import { Signal } from '@phosphor/signaling';
 
 export class DebuggerConsoleHandler implements IDisposable {
-  constructor(options: DebuggerNotebookHandler.IOptions) {
+  constructor(options: DebuggerConsoleHandler.IOptions) {
```
jtpio: Nice catch :+1:

JohanMabille:
Alright, then let's keep the `activate` methods of each plugin and refactor smaller parts of their implementation so we avoid some code duplication while keeping flexibility.

jtpio:
That sounds like a good compromise :+1: 

We can also fix the `DebuggerNotebookHandler` namespace in the console handler (https://github.com/jupyterlab/debugger/pull/78#discussion_r337012964) in a separate PR.

JohanMabille:
Ok, PR incoming shortly

JohanMabille:
@jtpio we can go further in refactoring old setting handler, but that would require function with constructor generic (the setOldHandler function should accept type like `new(a: any): T` with `T = DebuggerConsoleHandler | DebuggerNotebookHandler`).

JohanMabille:
@jtpio this is ready for review.

jtpio:
Thanks @JohanMabille.

cc @afshin who might want to look at it.

Issue #76, opened 2019-10-21
Handle kernel status


jtpio:
For now we don't check whether a kernel is ready before sending debug messages.

We should do that check before starting a new session and restoring an existing one. And also handle status changes.

jtpio:
Looking into that one.

jtpio:
We should also handle the case when switching to a kernel that does not support debugging (for example from `xeus-python` to `ipykernel`).

In that case the debugger UI will have to be removed (or at least disabled).

PR #82, opened 2019-10-22
Wait for the session to be ready before restoring state

jtpio:
Fixes part of #76:

- wait for the session to be ready before restoring the state
- throw an error is the kernel is not available when sending messages

This could be improved later, for example by automatically restoring the debug state when the kernel is restarted.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -116,7 +116,7 @@ export class DebugSession implements IDebugger.ISession {
 
       await this.sendRequest('attach', {});
     } catch (err) {
-      console.error('Error: ', err.message);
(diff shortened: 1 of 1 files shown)


JohanMabille:
Notice that restarting the kernel kills the debugger (so the initialize sequence + list of breakpoints should be sent back upon kernel restart).

jtpio:
Yes.

The first thing is that we need to use a signal such as `client.statusChanged` to know when a kernel was restarted (or dead, connecting...).

Then the state (in-memory breakpoints) would need to be retrieved from the editors.

PR #89, opened 2019-10-23
[WIP] Send breakpoints to the kernel on breakpointChanged

### LINKED ISSUE #43: Send breakpoints to the debugger
To be implemented after https://github.com/jupyterlab/debugger/issues/42.

Breakpoints set in the UI should be sent over to the debugger all at once (the debugger must be started beforehand).

jtpio:
Initial steps to send the breakpoints to the kernel every time they change in the UI (added or deleted).

This is still a work in progress as there seems to be some issues w.r.t the handlers (messages sent multiple times).

Fixes #43

### DIFF
--- a/src/breakpoints/index.ts
+++ b/src/breakpoints/index.ts
@@ -80,6 +80,9 @@ export namespace Breakpoints {
       this._breakpoints = model;
     }
 
+    breakpointsChanged = new Signal<this, IBreakpoint[]>(this);
+    clearedBreakpoints = new Signal<this, SessionTypes | null>(this);
+
     get breakpoints(): IBreakpoint[] {
       return this._breakpoints;
     }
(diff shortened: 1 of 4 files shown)


JohanMabille:
Can you explain how to reproduce the issue you had? I tried your branch and it seems that the breakpoints are sent correctly according to PTVSD logs.

jtpio:
multiple-messages-breakpoints

It's probably not related to this change. It looks like the handlers are not removed correctly when switching between notebooks and cells.

JohanMabille:
Is it ok for you if I merge so that I can iterate?

jtpio:
Sure.

We can log this (https://github.com/jupyterlab/debugger/pull/89#issuecomment-546677211) in a separate issue.

JohanMabille:
yep definitely

Issue #74, opened 2019-10-21
Resize Name and Value columns in the Variables table


jtpio:
Ideally we should be able to resize the "Name" and "Value" columns in the "Variables" table, to be able to read long variable names:

![image](https://user-images.githubusercontent.com/591645/67194377-4add4e80-f3f7-11e9-8829-034b69887338.png)

jtpio:
Not something to tackle now, but to keep in mind for later.

We might actually consider using a tree instead of a table (mentioned in https://github.com/jupyterlab/debugger/issues/30). Using a tree would also make it easier to display classes (https://github.com/jupyterlab/debugger/issues/23).

KsavinN:
Done in #66

Issue #30, opened 2019-08-28
Consider using react-inspector for the variable explorer


jtpio:
The `react-inspector` library could be a nice tool to consider to show a tree of variables, that would handle the different scopes ("Locals", "Globals", "builtins") out of the box.

It's being used in the [jupyterlab-kernelspy](https://github.com/vidartf/jupyterlab-kernelspy) and the [jupyterlab-data-explorer](https://github.com/jupyterlab/jupyterlab-data-explorer/blob/a39a8afa955ff1ce4629d4739025bb8116f866b8/dataregistry-extension/src/debugger.tsx#L55_) extensions.

An example of what it would look like:

![image](https://user-images.githubusercontent.com/591645/63847359-ede09180-c98d-11e9-81ae-445c01c1805c.png)

The tree (headers, labels) can be customized so we could show the most important information at the higher levels (variable name and value). And end up with something similar to VS Code:

![image](https://user-images.githubusercontent.com/591645/63847873-fbe2e200-c98e-11e9-9bc9-236566c74838.png)

jtpio:
Quick example with labels and values:
 
![variable-react-inspector](https://user-images.githubusercontent.com/591645/63853201-5e8dab00-c99a-11e9-84a9-a7dbe7cac2fe.gif)

KsavinN:
Interesting didn't know about that. Thx :)

KsavinN:
> Quick example with labels and values:
> 
> ![variable-react-inspector](https://user-images.githubusercontent.com/591645/63853201-5e8dab00-c99a-11e9-84a9-a7dbe7cac2fe.gif)

What is that an extension ?

jtpio:
An early prototype for the debugger extension that we used to test the protocol and Codemirror apis (gutter marks and line highlighting)

KsavinN:
I guess a code still exist. @jtpio  Could you send me a link if of course repo still exist?

jtpio:
https://github.com/QuantStack/jupyterlab-debugger (mostly a quick proof of concept)

All development should now happen in the [present repo](https://github.com/jupyterlab/debugger) :)

Maybe we can open a PR later once we have the integration between the UI and the kernel via the DAP (Debug Adapter Protocol) so we can try that `react-inspector` out.

afshin:
Sweet, I didn't realize the react-inspector existed, this is useful.

KsavinN:
Resolved in #91

PR #91, opened 2019-10-23
Implement new visualization of variables

KsavinN:
chrome-capture (4)

Related with:
#30 
#74


src/variables/body/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  const List = () => (
+    <>
+      {data.map(scopes => (
+        <ObjectInspector
+          key={scopes.name}
+          data={scopes.variables}
+          name={scopes.name}
+          nodeRenderer={defaultNodeRenderer}
```
jtpio: we also need to pass the `theme` prop that uses the JupyterLab CSS values to support the dark theme

KsavinN: good point

jtpio:
Would it be possible to name the commits with a small description of the modifications, so it's easier to know what has changed?

For example, https://github.com/jupyterlab/debugger/pull/91/commits/cdfd31957b96e72bf2ab731069ccffd09b043dfd adds a default `[]` for the scopes but the commit message is `Merge branch 'UiTree' of github.com:KsavinN/debugger into UiTree`.

src/variables/body/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  return <>{List()}</>;
+};
+
+const theme = {
+  BASE_FONT_FAMILY: 'var(--jp-code-font-family)',
+  BASE_FONT_SIZE: 'var(--jp-code-font-size)',
+  BASE_LINE_HEIGHT: 'var(--jp-code-line-height)',
+
+  BASE_BACKGROUND_COLOR: 'var(--jp-layout-color0)',
```
jtpio: Shouldn't it this color instead?

```suggestion
  BASE_BACKGROUND_COLOR: 'var(--jp-layout-color1)',
```

image

src/variables/body/search.tsx
```diff
@@ -16,14 +16,11 @@ export class Search extends Widget {
 
     const layout = new PanelLayout();
     this.layout = layout;
-    this.scope = new ScopeSearch(model);
     this.search = new SearchInput(model);
 
-    layout.addWidget(this.scope);
-    layout.addWidget(this.search);
+    // layout.addWidget(this.search);
```
jtpio: This comment can be removed

typings/react-inspector/react-inspector.d.ts
```diff
@@ -0,0 +1,181 @@
+// Type definitions for react-inspector
+// Project: https://github.com/xyc/react-inspector
+// Definitions by: vidartf <https://github.com/vidartf>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
```
jtpio: We can also mention they come from https://github.com/vidartf/jupyterlab-kernelspy

jtpio:
> Sure :)

Nice, thanks :+1:

jtpio:
It would also be nice if the first level of the tree could be expanded by default:

image


Instead of:

image

src/variables/body/search.tsx
KsavinN: I actually should remove that file cause with react-inspector we don't need search. (for now)

jtpio: Yes good point.

jtpio:
Thanks @KsavinN for this.

I pushed a minor color fix for the variable tree (last commit).

We can spend more time on this later, depending on whether we stick to react-inspector or not (see https://github.com/jupyterlab/debugger/issues/95).

jtpio:
Also removed the `filter` which isn't used anymore (at least for now).

PR #92, opened 2019-10-23
Remove cyclic reference sidebar <-> model

jtpio:
The debugger model is passed to the sidebar but not used.

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -37,7 +37,7 @@ export class Debugger extends SplitPanel {
 
     this.model = new Debugger.Model(options);
 
-    this.sidebar = new DebuggerSidebar(this.model);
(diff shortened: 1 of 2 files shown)


jtpio:
@KsavinN @JohanMabille @afshin small cleanup ready for review

JohanMabille:
Indeed only the parts of the UI that update data should have a reference on the model.

afshin:
> Indeed only the parts of the UI that update data should have a reference on the model.

Do you consider this a hard rule? I would have considered some widgets that treat a model as read-only still acceptable. What do you think?

JohanMabille:
Actually your remark made me realize that my sentence was incomplete, since we may have views that need to display data only (without updating them); in that case they would need a reference on a model, and I agree with you that a readonly reference would be the best option.

Issue #90, opened 2019-10-23
Architecture of the debugger


JohanMabille:
Nothing blocking at all, these are just some questions and thoughts about how internal components should interact.

- Should we hide the DebugSession behind the DebugService? In this case, the DebugService would provide methods to set the session, start and stop it.
- Should the DebugService only provide methods to send debug request and react to debug events? In this case, it would be responsible for updating the "data" of different model classes (Breakpoints, Variables).
- Should the debugger model hold other model classes (breakpoints, variables)?

cc @afshin @jtpio @KsavinN

jtpio:
> Should we hide the DebugSession behind the DebugService? In this case, the DebugService would provide methods to set the session, start and stop it.

> Should the DebugService only provide methods to send debug request and react to debug events? In this case, it would be responsible for updating the "data" of different model classes (Breakpoints, Variables).

We might not need the service actually. The model would expose higher level method such as `setBreakpoints` or `getVariables`, and would update its state on response.

Still using its `DebugSession` for sending messages to the kernel.

>  Should the debugger model hold other model classes (breakpoints, variables)?

With `model` being a `DebuggerModel`, `model.breakpoints` a `Breakpoints.Model` and so on.

UI elements could update their states on `model.breakpointsChanged` and other signals.

JohanMabille:
> We might not need the service actually. The model would expose higher level method such as setBreakpoints or getVariables, and would update its state on response

Actually I would like to keep the model as simple as possible, basically data (or other data models as you described) and signals, and implement the logic in a dedicated class.

JohanMabille:
So basically we have the following options:

1] The service has a reference on the model which only contains data / other data models and signals that emit when the data change. The service is connected to these signals so he can send the appropriate requests to the kernel. The model does not know the service.

2] The model owns the service and connects its slots to the signals of the service (which will basically only forward debug events). When data is changed, the model calls the appropriate service method.

3] The service is directly implemented in the model. It is equivalent to 2] but simpler (although more monolithic). In that case we definitely need to expose an IModel interface if we want to be able to drive the debugger from other plugins / parts of the app.

#93 implements (partially) the first option, which is better regarding the principle of separation of concerns. The drawbacks is that it slightly complicates the workflow, however, if we keep dedicated models for breakpoints and callstacks, we will have to connect slots to their data changed signal anyway (at least to notify the main model).

JohanMabille:
Here is a overview of the target architecture:

[...]

PR #93, opened 2019-10-24
Added interface for service, moved session into service

JohanMabille:
This is a slight refactoring along the lines of ideas exposed in #90. Things that will be done if this gets accepted:

- remove the dependency on IService from the Notebook, Console and Cell handlers (I added this dependency for compilation purpose only).
- complete IService interface and replace DebugService type with IDebug.IService type in the debugger.
- maybe expose an IModel interface in token.ts? Not sure about this.

The global idea is that implementation classes should only rely on interfaces (except where concrete classes are implemented). Exposing an IService interface would allow to drive a debugger launched in a notebook form a external console for instance (which could be really cool!)


src/service.ts
```diff
@@ -10,37 +10,50 @@ import { Variables } from './variables';
 
 import { Callstack } from './callstack';
 
-export class DebugService {
-  constructor(session: DebugSession | null, debuggerModel: Debugger.Model) {
-    this.session = session;
+export class DebugService implements IDebugger.IService {
+  constructor(debuggerModel: Debugger.Model) {
+    this._session = null;
```
jtpio: Should we keep the previous comment about keeping the session to `null` here too?

JohanMabille: Ah good catch!

Issue #85, opened 2019-10-22
Show current line


jtpio:
Editors should show the line that corresponds to where the execution has stopped.

Something similar to this:

![image](https://user-images.githubusercontent.com/591645/67298887-2d3de100-f4ec-11e9-8c71-e3642cc2b875.png)

This information can be retrieved from the `StackFrame`: https://microsoft.github.io/debug-adapter-protocol/specification#Types_StackFrame

jtpio:
Fixed in #94.

Issue #86, opened 2019-10-22
Add `continue` command


jtpio:
There should be a new command for the `continue` debug action, so it can be mapped to a keyboard shortcut (F5 in regular IDEs, F8 in the dev tools).

PR #94, opened 2019-10-24
Add line highlight for breakpoints

KsavinN:
chrome-capture (5)

resolved:
#85

In future:
#86  

I add simple Highline. Signal instance is created in the model of the debugger. Emitting is placed in service and signal is connected in cell handler. Is that proper way or should we consider about another architecture?


src/debugger.ts
```diff
@@ -172,6 +176,7 @@ export namespace Debugger {
     private _session: IDebugger.ISession | null;
     private _sessionChanged = new Signal<this, void>(this);
     private _service = new DebugService(null, this);
+    private _selectCurrentLine = new Signal<this, number>(this);
```
jtpio: How about this name to follow the other conventions?

```suggestion
    private _currentLineChanged = new Signal<this, number>(this);
```

style/breakpoints.css
```diff
@@ -13,3 +13,7 @@
   top: -1px;
   color: #f22;
 }
+
+.highlight {
```
jtpio: We could think of a more descriptive class name.

style/breakpoints.css
```diff
@@ -13,3 +13,7 @@
...
+  background-color: orange;
```
jtpio: Here we could make use of a JupyterLab CSS variable (to also handle the dark theme)

jtpio:
> Emitting is placed in service and signal is connected in cell handler. Is that proper way or should we consider about another architecture?

I think it makes sense to have it there for now and let the service update the model.

Later the service will be able to:

1. listen to `stopped` events
2. retrieve `scopes`, `stackTraces` and `variables` on `stopped`
3. update the model data (`scopes`, `variables`, `currentLine` and so on)

The views would listen to changes on their model and reflect the updates.

jtpio:
Note that this change doesn't remove the line highlighting when the debugger stops.

We'll need some kind of cleanup handler for that.

src/handlers/cell.ts
```diff
@@ -36,6 +43,24 @@ export class CellManager implements IDisposable {
   private _activeCell: CodeCell;
   isDisposed: boolean;
 
+  showCurrentLine(lineNumber: number) {
```
jtpio: Could this method actually be private?

KsavinN:
I added a suggestion of @jtpio.
If everything looks ok I guess we can merge that.
Rest development will be go on after PR #93.
@afshin

jtpio:
Going to rebase this one to fix the conflicts.

jtpio:
After getting the latest changes from master (tree view):

line-highlight

KsavinN:
> After getting the latest changes from master (tree view):
> 
> One more step to real debugger :)

PR #98, opened 2019-10-25
Proxy

JohanMabille:
This splits the Debugger class in two parts, one (Debugger) containing the service and model and one pure Widget (DebuggerWidget) containing the UI elements and the Debugger.

The Debugger instance is always available and can replace the proxy object in the activate method of the main plugin. It is passed to the DebuggerWidget upon construction.

Notice that this is not finished as an exception is raised whenever a notebook or console is created while the debugger widget has not been created yet. This is because the different handlers try to access breakpoints with `model.sidebar.breakpoints.mode` while the sidebar has not been created yet.

This makes me think that we (I) might have merged #92 too fast and that the opposite should be done: the sidebar has a reference on the main model, the main model class has references on other models (breakpoints, callbacks etc) so they are always available for the handlers.

EDIT: this PR is based on the commit from #97 to illustrate how ALL the isEnabled functions of commands are simpler now, it will be rebased once #97 is merged.

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -27,18 +27,42 @@ import { IDebugger } from './tokens';
 
 import { IDataConnector } from '@jupyterlab/coreutils';
 
-export class Debugger extends SplitPanel {
+export class Debugger implements IDebugger {
   constructor(options: Debugger.IOptions) {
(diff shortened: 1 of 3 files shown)


JohanMabille:
Superseded with #100

PR #99, opened 2019-10-25
Develop highlights

KsavinN:
chrome-capture (9)


jtpio:
In the screencast, why does the line jump to line 2 at some point?

src/debugger.ts
```diff
@@ -146,6 +151,7 @@ export namespace Debugger {
     private _mode: IDebugger.Mode;
     private _modeChanged = new Signal<this, IDebugger.Mode>(this);
     private _currentLineChanged = new Signal<this, number>(this);
+    private _clearLines = new Signal<this, void>(this);
```
jtpio: It would be great to keep the same naming convention for all the signals:

```suggestion
    private _linesCleared = new Signal<this, void>(this);
```

src/index.ts
```diff
@@ -281,7 +281,7 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
 
         sidebar.id = 'jp-debugger-sidebar';
         sidebar.title.label = 'Environment';
-        shell.add(sidebar, 'right', { activate: false });
+        shell.add(sidebar, 'right', { activate: true });
```
jtpio: Do we always want to activate it?

src/service.ts
```diff
@@ -134,7 +127,11 @@ export class DebugService implements IDebugger.IService {
 
     this.session.client.kernel.requestExecute({ code });
 
-    const stackFrames = await this.getFrames(threadId);
+    await this.getFramesAllData();
+  }
+
+  getFramesAllData = async () => {
```
jtpio: How about `getAllFrames` instead?

KsavinN:
> In the screencast, why does the highlight jump to line 2 at some point?

It seems in StackFrame is line: 2. I will check that

jtpio:
Interesting. It could have been because of the empty lines at the beginning being stripped away.

src/debugger.ts
```diff
@@ -146,6 +151,7 @@ export namespace Debugger {
     private _mode: IDebugger.Mode;
     private _modeChanged = new Signal<this, IDebugger.Mode>(this);
     private _currentLineChanged = new Signal<this, number>(this);
+    private _LinesCleared = new Signal<this, void>(this);
```
jtpio: The capital "L" looks like a type:

```suggestion
    private _linesCleared = new Signal<this, void>(this);
```

src/service.ts
```diff
@@ -110,15 +112,14 @@ export class DebugService implements IDebugger.IService {
...
-        threadId = msg.body.threadId;
-      }
-    });
+
+    // for now we dont need that -> need develop that
+    // this.session.eventMessage.connect((_, event: IDebugger.ISession.Event) => {
+    //   const eventName = event.event;
+    //   if (eventName === 'thread') {
+    //   }
+    // });
```
jtpio: So maybe we can remove it from the code base for now?

jtpio:
Trying this out locally, I am seeing something similar to the example above:

offset-line-stopped

KsavinN:
> Trying this out locally, I am seeing something similar to the example above:
> 
> offset-line-stopped

A breakpoint is set for variable which is set for creating instance. The first stackFrames is probably takie part in class A.

jtpio:
Yes, but we want to highlight the `a = A()` line here.

KsavinN:
> Yes, but we want to highlight the `a = A()` line here.

For know, highlights have input from stackFrames. I will check if we need to implement some think more to get right highlight;

jtpio:
So the first frame returned is indeed with line 4 for this example. And the highlighted line corresponds to the line from the first frame.

jtpio:
OK I looked at the VS Code logs to compare and dev tools WS messages.

This is the message we send to the kernel after adding a breakpoint on line 6:

```json
"arguments": {
      "breakpoints": [
        {
          "line": 5
        }
      ],
      "source": {
        "path": "/tmp/xpython_27949/2552399841628451373.py"
      },
      "sourceModified": false
    }
```

This is part of the `debug_reply` message returned by the kernel:

```json
"body": {
      "breakpoints": [
        {
          "id": 1,
          "line": 4,
          "source": {
            "path": "/tmp/xpython_27949/2552399841628451373.py"
          },
          "verified": true
        }
      ]
    },
```

In VS code if you set a breakpoint on line 5 it actually "jumps" to line 4. So I suspect this is what is happening here (off by one when setting the breakpoints).

jtpio:
That would explain why the execution stops on line 4.

src/service.ts
```diff
@@ -198,7 +203,7 @@ export class DebugService implements IDebugger.IService {
   getBreakpoints = (): DebugProtocol.SourceBreakpoint[] => {
     return this._model.sidebar.breakpoints.model.breakpoints.map(breakpoint => {
       return {
-        line: breakpoint.line
+        line: breakpoint.line - 1
```
jtpio: This seems to be the reason for https://github.com/jupyterlab/debugger/pull/99#issuecomment-546876211

jtpio:
Thanks @KsavinN!

I pushed a small cleanup commit. Also the line highlight behaves as expected, let's merge.

Issue #80, opened 2019-10-21
Access to the debugger widget across plugins


jtpio:
At the moment the debugger widget is accessed from the `notebooks` and `consoles` plugins via the `tracker` exposed from the [proxy](https://github.com/jupyterlab/debugger/blob/b15c240d9b13a42b71b92e6988bcde3861190eb5/src/index.ts#L403-L407).

A side effect of accessing the widget this way is code that doesn't look natural. Here `debug.session` is tested twice:

https://github.com/jupyterlab/debugger/blob/b15c240d9b13a42b71b92e6988bcde3861190eb5/src/index.ts#L185-L193

We should probably keep the proxy as a way to access the (single) instance of the debugger widget (similar to the `inspector` and `completer` extensions), but it looks we could change some of the values it proxies?

jtpio:
cc @afshin 

This is related to what we discussed today during the call.

jtpio:
Fixed by https://github.com/jupyterlab/debugger/pull/100

PR #100, opened 2019-10-26
Model now owns Breakpoints, Variables and callstack models

JohanMabille:
- [X] Make the `Debugger.Model` object own other models (Breakpoints, Variables, Callstack)
- [x] Clean Model constructors and attributes
- [x] Merge IService and IDebugger
- [x] Make the service object always available and clean the initialization of the pluggin

This PR removes the proxy objects and replaces it with a DebugService object inheriting from IDebugger. This object is always available and does not own anything. Its `model` property is set upon construction of the debugger (and reset to null when the debugger is disposed to ensure the model data will be released). Other handlers are now passed the service object only and retrieve the debugger model thanks to this object.

Notice that the interface (i.e. IDebugger) provides `model` as a readonly property.

This suggest that we should expose an IModel interface from witch the model inherits, but this requires further refactoring (we would need to expose interface for breakpoints, callstack and variables model). I think it should be done in another PR to avoid too many conflicts.


jtpio:
Thanks @JohanMabille.

This needs to be rebased now that #99 is merged.

src/debugger.ts
```diff
@@ -21,43 +17,51 @@ import { ISignal, Signal } from '@phosphor/signaling';
...
 
 export class Debugger extends SplitPanel {
   constructor(options: Debugger.IOptions) {
     super({ orientation: 'horizontal' });
     this.title.label = 'Debugger';
     this.title.iconClass = 'jp-BugIcon';
 
-    this.model = new Debugger.Model(options);
+    this.model = new Debugger.Model({
+      connector: options.connector
```
jtpio: Not related to this change, but I'm wondering if we actually still need the connector here.

The `DebugSession` is the one acting as the connector when comparing it to the JupyterLab `inspector` extension.

jtpio: We can track that in a separate issue though.

jtpio:
This will fix https://github.com/jupyterlab/debugger/issues/80

src/debugger.ts
jtpio: Opened https://github.com/jupyterlab/debugger/issues/104

PR #105, opened 2019-10-28
Cleanup frames and variables on continued event

jtpio:
Cleanup frames and variables when the execution has terminated.

exited-event

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -70,7 +70,7 @@ export class DebugService implements IDebugger {
         void this.getAllFrames();
       } else if (event.event === 'continued') {
         this._threadStopped.delete(event.body.threadId);
-        this._model.linesCleared.emit();
+        this.onContinued();
       }
       this._eventMessage.emit(event);
     });
@@ -255,6 +255,12 @@ export class DebugService implements IDebugger {
       };
     });
   };
+
+  private onContinued() {
+    this._model.linesCleared.emit();
+    this._model.callstackModel.frames = [];
+    this._model.variablesModel.scopes = [];
+  }
 
   private currentThread(): number {
     // TODO: ask the model for the current thread ID


jtpio:
To be rebased on #100.

jtpio:
Also the `terminated` or `exited` events don't seem to be sent when the execution of the cell has finished (looking into it).

JohanMabille:
If I understand the protocol correctly, the `exited` event is triggered when the debuggee has exited and returned its exit code. However, here the debuggee is the kernel itself (ptvsd is started by making the kernel executing `import ptvsd ...`) so this event cannot be triggered while the kernel is alive.

jtpio:
Yes that would explain why.

jtpio:
The same goes for `terminated`.

JohanMabille:
I thought `terminated` meant the debugger had been stopped (so we should receive it when executing the `stop` command.

src/service.ts
```diff
@@ -240,6 +240,12 @@ export class DebugService implements IDebugger.IService {
     });
   };
 
+  private onContinued() {
+    this._model.linesCleared.emit();
+    this._model.sidebar.callstack.model.frames = [];
+    this._model.sidebar.variables.model.scopes = [];
```
JohanMabille: This should be replaced with
```typescript
    this._model.callstackModel.frames = [];
    this._model.variablesModel.scopes = [];
```

jtpio: Yes indeed, first rebase wasn't on #100 :)

jtpio:
This is now rebased on #100.

Instead of handling the `exited` and `terminated` events, we can decide to always clear the state (breakpoints and frames for now) on `continued`.

The new state will be populated on `stopped`. If there is no more `stopped`event (cell execution has finished), the UI will already be cleared.

jtpio:
Also it makes to clear the state when there is for example a long computation between 2 breakpoints (as pointed by @JohanMabille). This is also what is happening in VS Code, for example with (breakpoints on the two print statements):

```python
import time
print('foo')
time.sleep(10)
print('bar')
```

PR #112, opened 2019-10-28
Clear the model when the debugger is stopped

### LINKED ISSUE #110: Handle stop
#84 handles when the execution of a cell has `continued`.

But for now if we click on stop the variables and callstack are still shown:

stop-debugger-cleanup

JohanMabille:
Fixes #110

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -70,7 +70,7 @@ export class DebugService implements IDebugger {
         void this.getAllFrames();
       } else if (event.event === 'continued') {
         this._threadStopped.delete(event.body.threadId);
-        this.onContinued();
+        this.clearCallStack();
       }
       this._eventMessage.emit(event);
     });
@@ -87,6 +87,15 @@ export class DebugService implements IDebugger {
 
   isThreadStopped(): boolean {
     return this._threadStopped.has(this.currentThread());
+  }
+
+  async start(): Promise<void> {
+    await this.session.start();
+  }
+
+  async stop(): Promise<void> {
+    await this.session.stop();
+    this.clearModel();
   }
 
   async continue(): Promise<void> {
@@ -256,10 +265,16 @@ export class DebugService implements IDebugger {
     });
   };
 
-  private onContinued() {
+  private clearCallStack() {
     this._model.linesCleared.emit();
     this._model.callstackModel.frames = [];
     this._model.variablesModel.scopes = [];
+  }
+
+  private clearModel() {
+    this.clearCallStack();
+    this._model.breakpointsModel.breakpoints = [];
+    this._threadStopped.clear();
   }
 
   private currentThread(): number {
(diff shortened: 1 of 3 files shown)


src/service.ts
```diff
@@ -262,6 +271,12 @@ export class DebugService implements IDebugger {
     this._model.variablesModel.scopes = [];
   }
 
+  private clearModel() {
+    this.onContinued();
```
jtpio: This `onContinued` feels a bit off in the `clearModel` method.

Maybe we can move its content to `clearModel`? (or rename it)

JohanMabille: We cannot move it inside `clearModel` since `onContinue` is not supposed to clear the breakpoints.

jtpio: :+1:

src/service.ts
```diff
@@ -256,12 +265,18 @@ export class DebugService implements IDebugger {
...
-  private onContinued() {
+  private clearCallStack() {
     this._model.linesCleared.emit();
     this._model.callstackModel.frames = [];
     this._model.variablesModel.scopes = [];
   }
 
+  private clearModel() {
+    this.clearCallStack();
+    this._model.breakpointsModel.breakpoints = [];
```
jtpio: The breakpoints are removed from the model but not (yet) in the UI?

JohanMabille: Fix incoming ;)

afshin: * Doesn't setting that `[]` automatically trigger the `breakpointsChanged` signal?
* Do we needo *both* `breakpointChanged` and also `breakpointsChanged`? I think if all the models just had a simple `changed` value and they re-rendered it might be okay, because it would be the virtual DOM's responsibility to re-render efficiently, maybe. What do you think?

JohanMabille: My understanding of the workflow is the following:
- when clicking on the gutter to set a breakpoint, the setBreakpoints method of the model is called
- this last one triggers a `breakpointsChanged` signal
- the cell handler listens to this signal and calls the `updateBreakpoints` on the service object, which sends a debug request to the kernel to set the new breakpoints.

What we want to do here is the opposite: updating the gutter based on changes of the model.

Regarding the signals I don't know what is the most efficient option, but I agree that it would be better if we could keep only one signal.

JohanMabille: Do you think it would be better not to set the gutter markers in `onGutterClick` but rather when the cell handler receives a `breakpointsChanged` signal?

jtpio: > Doesn't setting that `[]` automatically trigger the `breakpointsChanged` signal?

I think that's the case yes.

Agree there should be only one `breakpointsChanged :+1:

JohanMabille: Actually I wonder if we should not update the gutter based on the breakpointsChanged event only, and adopt this behavior for other parts of the UI:
- a user's input calls some methods of the service, but we do not update parts of the view that depends on data of the model.
- the service sends a request to the backend if that's needed, and updates the mdel
- updating the model triggers a signal which forces the UI to rerender.

PR #113, opened 2019-10-29
Resolve issue with notebook handler

### LINKED ISSUE #102: Handlers issue
chrome-capture (10)

Now the issue look like that :
breakpoints_issue

KsavinN:
resolve #102


src/handlers/notebook.ts
```diff
@@ -46,28 +48,33 @@ export class DebuggerNotebookHandler implements IDisposable {
...
-    } else {
-      this.cellManager = new CellManager({
-        breakpointsModel: this.breakpoints,
-        activeCell: codeCell,
-        debuggerModel: this.debuggerModel,
-        debuggerService: this.debuggerService,
-        type: 'notebook'
-      });
-    }
+    // need timeout casue after dispose apear bad render problem
```
jtpio: What does "bad render problem" mean here? Maybe a quick screencast could be useful to show the issue?

KsavinN: Sorry, I didn't catch gif or video for that. After close notebook and go to first open notebook cell have to render box for line numbers but without number lines

jtpio: OK I'll have a look.

It's usually easier to track what is happening when connecting on signals rather than relying on `setTimeout`.

KsavinN: yeah, you have absolutely right. This is some hack resolving. I guess during close widget the signal of changing cell is faster than disposing of a old handler.

src/handlers/cell.ts
```diff
@@ -77,7 +76,6 @@ export class CellManager implements IDisposable {
     }
     this.removeListener(this.activeCell);
     this.cleanupHighlight();
-    Signal.clearData(this);
```
jtpio: Why do we need to remove this?

jtpio:
@KsavinN I tried this change locally without the `setTimeout` but didn't notice the render issue.

KsavinN:
> @KsavinN I tried this change locally without the `setTimeout` but didn't notice the render issue.

@jtpio  Thanks for check. I will try again on my local.

KsavinN:
@jtpio  I remove setTimeout and this happening :
chrome-capture (17)

jtpio:
OK. Is that real speed or low frame rate?

KsavinN:
> OK. Is that real speed or low frame rate?

real speed. Do you use white theme or dark for jupyter?

jtpio:
Both usually. I'll try to reproduce later.

jtpio:
What is the status on this PR? The diff panel on GH doesn't show any changes:

image

KsavinN:
It's still Work in Progress after merged new Iteration I found new bugs.

src/handlers/notebook.ts
```diff
@@ -46,28 +48,32 @@ export class DebuggerNotebookHandler implements IDisposable {
...
-        breakpointsModel: this.breakpoints,
-        activeCell: codeCell,
-        debuggerModel: this.debuggerModel,
-        debuggerService: this.debuggerService,
-        type: 'notebook'
-      });
-    }
+  protected onNewCell(noteTracker: NotebookTracker, codeCell: CodeCell) {
+    requestAnimationFrame(() => {
+      if (noteTracker.currentWidget.id === this.id) {
```
jtpio: What this means is that we would get the `activeCellChanged` signal here, sent from the notebook tracker, but the cell is from another notebook widget.

And the `cellManager` is created for one notebook widget.

src/handlers/notebook.ts
```diff
@@ -46,28 +48,32 @@ export class DebuggerNotebookHandler implements IDisposable {
...
+        if (this.cellManager) {
+          this.cellManager.activeCell = codeCell;
+        } else {
+          this.cellManager = new CellManager({
+            breakpointsModel: this.breakpoints,
+            activeCell: codeCell,
+            debuggerModel: this.debuggerModel,
+            debuggerService: this.debuggerService,
+            type: 'notebook'
```
KsavinN: And actually I'm considering is it exist some situation when cellManger is Created in this place.

KsavinN:
This PR is too much behind from master. Rest issues were resolved before

jtpio:
Yes it was lagging a bit behind.

We're still seeing the handler issues on master, so this should still be relevant. Maybe we can open a new PR since the diff isn't too big.

KsavinN:
> Yes it was lagging a bit behind.
> 
> We're still seeing the handler issues on master, so this should still be relevant. Maybe we can open a new PR since the diff isn't too big.

I think the new PR will be better.

Issue #103, opened 2019-10-28
Update the breakpoint locations


jtpio:
As mentioned in https://github.com/jupyterlab/debugger/pull/99 and https://github.com/jupyterlab/debugger/pull/99#issuecomment-546876211, the breakpoints actually set by the debugger might not correspond to the ones set in the UI.

For example this is the case when setting a breakpoint on an empty line.

We should read the `debug_reply` message received after `setBreakpoints` to retrieve the correct list of breakpoints, and update the breakpoint marks in the editor UI (and the line in the sidebar).

As an example in VS code when the debugger is running:

![vs-code-breakpoint-location](https://user-images.githubusercontent.com/591645/67674798-c605ae00-f97d-11e9-8e95-4eda977fe6d8.gif)

jtpio:
@JohanMabille we might want to do this one after https://github.com/jupyterlab/debugger/pull/129 and before https://github.com/jupyterlab/debugger/issues/130.

At the moment the `breakpointsModel` is directly modified by the cell handler, but this should be done by the service now. Otherwise we'll get two parties updating the model:
- the cell handler (i.e clicking on the gutter)
- the service to update modify the location of the breakpoints on the `setBreakpoints` response

This looks like the following logic should be removed or inverted (the service will update the breakpoints in the model, and the UI will react to the changes with the `breakpointsChanged` signal):

https://github.com/jupyterlab/debugger/blob/d1c0d9c88a654aaf91123a8c4c40f1756ab1e243/src/handlers/cell.ts#L44-L46

Issue #101, opened 2019-10-27
Breakpoints messages sent multiple times


JohanMabille:
As described in #89:

![multiple-messages-breakpoints](https://user-images.githubusercontent.com/591645/67632532-cbdb9080-f8a4-11e9-910f-64e6abf454b8.gif)

It looks like the handlers are not removed correctly when switching between notebooks and cells.

jtpio:
This seems to be fixed on master now (couldn't reproduce locally).

jtpio:
I haven't been able to reproduce it on master.

Some recent PRs have helped fixing that, so let's consider it as fixed for now.

PR #114, opened 2019-10-29
Keep line numbers on when switching cells

### LINKED ISSUE #106: Always show line numbers and gutters when the sidebar is open
At the moment the gutters and line numbers are being toggled when switching between cells:

codemirror-gutters-toggled

It would be better to always keep them on when the debugger sidebar is visible and being used.

We also want to keep the breakpoints set for other cells.

jtpio:
Fixes #106.

### TODO

- [x] Keep line numbers on when switching cells and the debugger sidebar is available
- [x] Do not clear breakpoints when switching cells
- [x] Retrieve existing breakpoints from the cell editor

edit: adding a screencast

multiple-cell-breakpoints

### DIFF
--- a/src/handlers/cell.ts
+++ b/src/handlers/cell.ts
@@ -119,7 +119,6 @@ export class CellManager implements IDisposable {
     ) {
       if (this.previousCell && !this.previousCell.isDisposed) {
         this.removeListener(this.previousCell);
-        this.clearGutter(this.previousCell);
         this.breakpointsModel.breakpoints = [];
       }
       this.previousCell = this.activeCell;
--- a/src/handlers/cell.ts
+++ b/src/handlers/cell.ts
@@ -135,6 +135,15 @@ export class CellManager implements IDisposable {
 
     this.previousLineCount = editor.lineCount;
 
+    const editorBreakpoints = this.getBreakpointsInfo(cell);
+    editorBreakpoints.forEach(info => {
+      this.breakpointsModel.addBreakpoint(
+        this._debuggerService.session.client.name,
+        this.getEditorId(),
+        info
+      );
+    });
+
     editor.setOption('lineNumbers', true);
     editor.editor.setOption('gutters', [
       'CodeMirror-linenumbers',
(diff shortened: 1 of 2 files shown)


jtpio:
Current status:

keep-breakpoints-gutter

The breakpoints should be picked up from the editors to they can be displayed in the sidebar.

jtpio:
multiple-cell-breakpoints

jtpio:
Still seeing https://github.com/jupyterlab/debugger/issues/101 when switching cells.

jtpio:
I think this is ready to go.

The related issues to tackle after would be:

- https://github.com/jupyterlab/debugger/issues/119
- https://github.com/jupyterlab/debugger/issues/103
- https://github.com/jupyterlab/debugger/issues/101

src/handlers/cell.ts
```diff
@@ -136,6 +135,15 @@ export class CellManager implements IDisposable {
...
     this.previousLineCount = editor.lineCount;
 
+    const editorBreakpoints = this.getBreakpointsInfo(cell);
+    editorBreakpoints.forEach(info => {
+      this.breakpointsModel.addBreakpoint(
+        this._debuggerService.session.client.name,
+        this.getEditorId(),
+        info
+      );
+    });
```
JohanMabille: I wonder if there should be a method for updating all the breakpoints in a single pass (to avoid sending many breakpoints requests to the kernel). Anyway, this is an optimization that we can do later.

src/handlers/cell.ts
```diff
@@ -238,3 +250,12 @@ export interface ILineInfo {
   /** Array of line widgets attached to this line. */
   widgets: any;
 }
+
+namespace Private {
+  export function createMarkerNode() {
```
JohanMabille: Do we need to export this function?

src/handlers/cell.ts
jtpio: Yes it would be better :+1: 

Also thinking about simplifying the `Breakpoints.Model` (there are 2 breakpoint setters at the moment)

src/handlers/cell.ts
jtpio: The function is exported but the namespace is not.

PR #116, opened 2019-10-29
Format all imports at once

jtpio:
With the new `ordered-imports` option added in https://github.com/jupyterlab/debugger/pull/72 and the pre-commit hook, we started having the imports sorted automatically on commit.

I would suggest we format them all at once now to minimize conflicts in the near future.

Also it seems that the pre-commit hook will need to be updated to only format the files that have been changed.


jtpio:
@JohanMabille @KsavinN @afshin fine if we get this in?

jtpio:
Closing as this is done gradually with each new PR.

PR #123, opened 2019-10-29
[WIP] Get more variables during expand variable

KsavinN:
It's not an appropriate method, more like a hack.

ObjectInpsector from react-inspector doesn't have props for handle expand click.


jtpio:
That looks interesting.

However the diff for this PR is the same as https://github.com/jupyterlab/debugger/pull/113.

jtpio:
Now it also includes the changes from https://github.com/jupyterlab/debugger/pull/113 it seems.

KsavinN:
> Now it also includes the changes from #113 it seems.

yes. Accidentally I pushed that.

PR #125, opened 2019-10-30
Display more variables during expand variable like Class

KsavinN:
It's not an appropriate method, more like a hack.

ObjectInpsector from react-inspector doesn't have props for handle expand click.

NEW:
newVar

OLD:


KsavinN:
@jtpio  this is new PR correlated with #123  but from the rebased branch with merge current master

src/service.ts
```diff
@@ -182,6 +183,23 @@ export class DebugService implements IDebugger {
     return reply.body.scopes;
   };
 
+  getVariable = async (_: any, variable: DebugProtocol.Variable) => {
+    if (!variable && variable.variablesReference !== 0) {
```
jtpio: Can the 2nd part of this expression be evaluated?

src/service.ts
```diff
@@ -150,6 +150,7 @@ export class DebugService implements IDebugger {
     }
 
     this._model.callstackModel.currentFrameChanged.connect(this.onChangeFrame);
+    this._model.variablesModel.variableExapnded.connect(this.getVariable);
```
jtpio: Typo:

```suggestion
    this._model.variablesModel.variableExpanded.connect(this.getVariable);
```

src/variables/body/index.tsx
```diff
@@ -28,11 +29,29 @@ const VariableComponent = ({ model }: { model: Variables.Model }) => {
   const [data, setData] = useState(model.scopes);
 
   useEffect(() => {
+    const convert = (scopes: Variables.IScope[]) => {
+      const convertet = scopes.map(scope => {
```
jtpio: converted typo?

```suggestion
      const converted = scopes.map(scope => {
```

jtpio:
Thanks @KsavinN for investigating this.

That might look like a workaround but maybe a good compromise for now while we stick to the react-inspector.

Will have a look at it locally.

src/service.ts
KsavinN: Some additional variable like in Class A (from gif) foo or bar have variable reference equal 0.
I guess that means we  cant get more description of this variable

jtpio: By that I mean:

- can `variable` be null or undefined?
- if so then isn't `variable.variablesReference` going to break?

jtpio:
Trying it out locally, I noticed the following behavior when expanding an int variable:

variable-expand-moreinfo

src/service.ts
KsavinN: I will check that

KsavinN:
> Trying it out locally, I noticed the following behavior when expanding an int variable:
> 
> ![variable-expand-moreinfo](https://user-images.githubusercontent.com/591645/68114744-12f60100-fef7-11e9-9afd-a86c2405ed8a.gif

@jtpio  Thanks for catching that. It probably correlated with merge master. I will fix that

src/variables/index.ts
```diff
@@ -96,17 +112,29 @@ export namespace Variables {
...
+
+    get variableExpanded(): ISignal<this, IVariable> {
+      return this._variableExpanded;
     }
 
     getCurrentVariables(): IVariable[] {
       return this.variables;
     }
 
+    getMoreDataOfVariable(variable: IVariable) {
```
jtpio: Maybe this could be called `expandVariable` if what it does is just emitting `_variableExpanded`?

KsavinN: Or can be onExpandVariable?

src/service.ts
```diff
@@ -328,6 +329,20 @@ export class DebugService implements IDebugger {
     return reply.body.scopes;
   };
 
+  getVariable = async (_: any, variable: DebugProtocol.Variable) => {
+    const reply = await this.session.sendRequest('variables', {
+      variablesReference: variable.variablesReference
+    });
+    let newVariable = { ...variable };
+    reply.body.variables.forEach((ele: DebugProtocol.Variable) => {
+      newVariable = { [ele.evaluateName]: ele, ...newVariable };
```
jtpio: Should this be `ele.name` instead of `ele.evaluateName`? Otherwise attributes are displayed `a.foo` instead of `foo` for the following code sample:

```python
class A:
    def __init__(self):
        self.foo = 'foo'
        self.bar = 'bar'

a = A()
```

jtpio: Also we could rename `ele` to`variable` to be more explicit.

src/service.ts
```diff
@@ -328,6 +329,20 @@ export class DebugService implements IDebugger {
...
+    });
+
+    this._model.variablesModel.currentVariable = newVariable;
```
jtpio: Do we need to keep track of the `currentVariable` in the model?

jtpio:
Thanks @KsavinN for the updates, I will have another look at it.

jtpio:
It looks like a lot of `variables` requests are being sent to the kernel, before expanding the variables:

variables-expand

jtpio:
The last commit changes the behavior slightly, so the `variables` request is only sent when the node is expanded:

expand-on-click

As a side-effect, there is a short period of time where we see the previous content, but I think it's fine for now. As a workaround we could check if it's possible to override the empty dict when the node is collapsed.

jtpio:
Looking at a few other details before merging (switching to wip).

src/variables/body/index.tsx
```diff
@@ -60,6 +113,24 @@ const VariableComponent = ({ model }: { model: Variables.Model }) => {
...
+  }
+  if (type === 'bool') {
+    return value === 'False' ? false : true;
+  }
+  if (type === 'str') {
+    return (value as string)
+      .split('')
+      .slice(1, value.length - 1)
+      .join('');
+  }
```
jtpio: I was wondering why we had to do that, but that's because strings are returned with single quotes:

image

jtpio: We should be able to call `.slice(1, value.length - 1)` directly on the string.

KsavinN: yeah you have right.

jtpio: I just pushed a small refactoring :+1:

jtpio:
OK I think we can stop here for now.

Thanks again @KsavinN, this is already a good improvement. We can open a new issue with the things to do as follow-up.

PR #126, opened 2019-10-30
Automagically starts the debugger if the kernel supports it

### LINKED ISSUE #117: Always start the debugger whenever possible?
Something we discussed together with @afshin, @KsavinN and @JohanMabille today.

We should probably always start the `DebugSession` whenever possible. That means when a kernel is available and that kernel supports debugging.

JohanMabille:
Fixes #117 and #110.

This removes the start and stop commands form the palette and implements the following behavior:
- when clicking on "Debugger" in the palette, the kernel of the notebook or console that has the focus starts a debugger if it can
- when opening a notebook or a console while the Debugger panel is already mounted, he associated kernel starts a debugger it it can.
- a running debugger is never manually stoppable (it will be stopped when the kernel dies).

### DIFF
--- a/src/breakpoints/index.ts
+++ b/src/breakpoints/index.ts
@@ -3,11 +3,11 @@
 
 import { Toolbar, ToolbarButton } from '@jupyterlab/apputils';
 
-import { Widget, Panel, PanelLayout } from '@phosphor/widgets';
-import { DebugProtocol } from 'vscode-debugprotocol';
-import { Body } from './body';
 import { Signal } from '@phosphor/signaling';
+import { Panel, PanelLayout, Widget } from '@phosphor/widgets';
+import { DebugProtocol } from 'vscode-debugprotocol';
 import { ILineInfo } from '../handlers/cell';
+import { Body } from './body';
 
 export class Breakpoints extends Panel {
   constructor(options: Breakpoints.IOptions) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -79,6 +79,9 @@ async function setDebugSession(
     debug.session.client = client;
   }
   await debug.session.restoreState();
+  if (debug.canStart()) {
+    await debug.start();
+  }
   app.commands.notifyCommandChanged();
 }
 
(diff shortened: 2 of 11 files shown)


JohanMabille:
This is ready for review and merge. There is a workaround due to breakpoints being updated upon each cell change. THis can lead to send a debug request to a kernel that is not ready yet.

A betteer fix would be to keep the list of all the breakpoints of a Notebook, and let the restoreSession update them when we switch to another notebook.

src/breakpoints/index.ts
```diff
@@ -92,11 +92,13 @@ export namespace Breakpoints {
     }
 
     set breakpoints(breakpoints: IBreakpoint[]) {
+      console.log('setbreakpoints called');
```
jtpio: Left over console.logs?

JohanMabille: oups ;)

src/index.ts
```diff
@@ -79,6 +79,9 @@ async function setDebugSession(
     debug.session.client = client;
   }
   await debug.session.restoreState();
+  if (debug.canStart()) {
```
jtpio: Wondering if this shouldn't be before `restoreState` (not sure yet).

jtpio:
Pretty neat.

Found this while trying it out locally:

restart-kernel-switch-cell

But it's probably something that falls under https://github.com/jupyterlab/debugger/issues/76.

src/index.ts
JohanMabille: I don't think so because you might restore a session that already has a running debugger.

JohanMabille:
Yes and it should be solved if we implement #130 (meaning we wouldn't have to check if the kernel is ready at all)

jtpio:
This fixes https://github.com/jupyterlab/debugger/issues/117 and https://github.com/jupyterlab/debugger/issues/110.

jtpio:
The other changes in the diff are related to the linting (https://github.com/jupyterlab/debugger/pull/116 and more will come).

PR #129, opened 2019-10-30
Set editors breakpoints on model changed

### LINKED ISSUE #119: Breakpoints on the gutter
Currently, clicking the gutter directly adds the breakpoint [marker](https://github.com/jupyterlab/debugger/blob/master/src/handlers/cell.ts#L177).

The correct behavior would be to add this marker when the model signals a change in its breakpoints.

jtpio:
Fixes #119 

The idea is to have one central place where breakpoints are added to the editor, that should be triggered when the model changes.

### Summary of the changes

- Keep breakpoints sorted by line number
- Unify the naming of the breakpoints model methods
- Pass line numbers instead of `ILineInfo` to the breakpoint model
- Remove `clearedBreakpoints` signal

### DIFF
--- a/src/breakpoints/index.ts
+++ b/src/breakpoints/index.ts
@@ -6,7 +6,6 @@ import { Toolbar, ToolbarButton } from '@jupyterlab/apputils';
 import { Signal } from '@phosphor/signaling';
 import { Panel, PanelLayout, Widget } from '@phosphor/widgets';
 import { DebugProtocol } from 'vscode-debugprotocol';
-import { ILineInfo } from '../handlers/cell';
 import { Body } from './body';
 
 export class Breakpoints extends Panel {
@@ -108,15 +107,7 @@ export namespace Breakpoints {
       }
     }
 
-    addBreakpoint(session: string, type: string, lineInfo: ILineInfo) {
-      const breakpoint: Breakpoints.IBreakpoint = {
-        line: lineInfo.line + 1,
-        active: true,
-        verified: true,
-        source: {
-          name: session
-        }
-      };
+    addBreakpoint(breakpoint: IBreakpoint) {
       this.breakpoints = [...this._breakpoints, breakpoint];
     }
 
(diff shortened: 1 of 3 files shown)


src/handlers/cell.ts
```diff
@@ -207,6 +202,19 @@ export class CellManager implements IDisposable {
     }
   };
 
+  private addBreakpointsToEditor(cell: CodeCell) {
+    this.clearGutter(cell);
+    const editor = cell.editor as CodeMirrorEditor;
+    const breakpoints = this._debuggerModel.breakpointsModel.breakpoints;
+    breakpoints.forEach(breakpoint => {
+      editor.editor.setGutterMarker(
+        breakpoint.line - 1,
```
jtpio: Not super happy of these `+/-1` lying around. Would be better to have a proper way to map that information.

jtpio:
Taking the opportunity to simplify the breakpoint model interface a little bit.

src/breakpoints/index.ts
```diff
@@ -81,7 +80,6 @@ export namespace Breakpoints {
     }
 
     breakpointsChanged = new Signal<this, IBreakpoint[]>(this);
-    clearedBreakpoints = new Signal<this, SessionTypes | null>(this);
```
jtpio: `clearedBreakpoints` is the same as `breakpointsChanged` for the special case of `[]`.

jtpio:
cc @JohanMabille if you wanted to base https://github.com/jupyterlab/debugger/issues/130 on these changes.

I've added a summary of the changes in the first comment.

There are still a few things we can improve, such as:

- Remove the manual `-1` when dealing with the line numbers. We should check if we could just get the line numbers with the right offsets directly from CodeMirror
- Remove the `breakpoint` setter and the `breakpointChanged` signal. At the moment they are used to handle the change of the `active` property in the `BreakpointComponent`.

PR #131, opened 2019-10-30
Hover on gutter

### LINKED ISSUE #127: Breakpoint markers hover in the gutters
It would be great to also support the hover effect on the gutter (before placing the breakpoint):

breakpoint-gutter-hover

KsavinN:
chrome-capture (18)

Resolve #127


src/handlers/cell.ts
```diff
@@ -186,6 +210,7 @@ export class CellManager implements IDisposable {
       'breakpoints',
       isRemoveGutter ? null : Private.createMarkerNode()
     );
+    setTimeout(this.setHover);
```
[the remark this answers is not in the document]
KsavinN: SetOptions on breakpoints remove all Wrapper gutter elements and set new. That process take a while (some milliseconds). that's why we need to use setTimeout. 

/Interesting. I would have expected CodeMirror to have something like:

editor.on('gutterHover', this.onGutterHover);
Similar to the click handler./

I will try that, I cant find in doc about gutterHover and my vs dont show that options

KsavinN:
> Interesting. I would have expected CodeMirror to have something like:
> 
> ```
> editor.on('gutterHover', this.onGutterHover);
> ```
> 
> Similar to the click handler.

I check the event for Editor and found  (scroll,blur,gutterClick, viewportChange,update,renderLine, overwriteToggle,beforeSelectionChange,cursorActivity,beforeChange)

src/handlers/cell.ts
```diff
@@ -10,12 +10,15 @@ import { Doc, Editor } from 'codemirror';
...
 import { IDisposable } from '@phosphor/disposable';
+
 import { Debugger } from '../debugger';
+
 import { IDebugger } from '../tokens';
 
 import { Signal } from '@phosphor/signaling';
 
 const LINE_HIGHLIGHT_CLASS = 'jp-breakpoint-line-highlight';
+const GUTTER_HOVER_CLASS = 'jp-breakpoint-hover';
```
afshin: Typically for transient classes, we use the `mod` convention, so I would suggest calling this `jp-mod-hover`.

src/handlers/cell.ts
```diff
@@ -144,6 +147,27 @@ export class CellManager implements IDisposable {
 
     editor.editor.on('gutterClick', this.onGutterClick);
     editor.editor.on('renderLine', this.onNewRenderLine);
+
+    this.setHover();
+  }
+
+  setHover() {
```
afshin: Is it impossible to do this work with pure CSS? Doing it with JS seems like overkill.

Issue #130, opened 2019-10-30
Keep all the breakpoints of a notebook in the model


JohanMabille:
The breakpoints model should keep the list of all the breakpoints of a notebook instead of those of the active cell only. Internally, whis can be achieved by storing a map of list, the keys of the map being the hash of a cell computed with the same method as the one for duming cell in the backend.

This has many advantages:
- no need to resend the breakpoints to the backend when we switch from one cell to another
- no need to reset the breakpoints to an empty list and then retrieve the breakpoints of the new active cell from the editor
- breakpoints are restored thanks to the restore service when the user switches to a different notebook / console.

Besides limiting the number of messages that are sent (which is not a prformance issue but rather a debugability issue), it would simplify the internal logic.

PR #132, opened 2019-10-30
Set breakpoints via the service

### LINKED ISSUE #103: Update the breakpoint locations

jtpio:
Fixes #103 

update-breakpoint-locations

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -234,15 +234,21 @@ export class DebugService implements IDebugger {
       sourceBreakpoints,
       dumpedCell.sourcePath
     );
-    this._model.breakpointsModel.breakpoints = reply.body.breakpoints.map(
-      breakpoint => {
-        return {
-          ...breakpoint,
-          active: true,
-          source: { path: this.session.client.name }
-        };
-      }
+    let kernelBreakpoints = reply.body.breakpoints.map(breakpoint => {
+      return {
+        ...breakpoint,
+        active: true,
+        source: { path: this.session.client.name }
+      };
+    });
+
+    // filter breakpoints with the same line number
+
+    kernelBreakpoints = kernelBreakpoints.filter(
+      (breakpoint, i, arr) =>
+        arr.findIndex(el => el.line === breakpoint.line) === i
     );
+    this._model.breakpointsModel.breakpoints = kernelBreakpoints;
     await this.session.sendRequest('configurationDone', {});
   };
 
--- a/src/service.ts
+++ b/src/service.ts
@@ -207,14 +207,6 @@ export class DebugService implements IDebugger {
     return reply.body.variables;
   };
 
-  getBreakpoints = (): DebugProtocol.SourceBreakpoint[] => {
-    return this._model.breakpointsModel.breakpoints.map(breakpoint => {
-      return {
-        line: breakpoint.line
-      };
-    });
-  };
-
   setBreakpoints = async (
     breakpoints: DebugProtocol.SourceBreakpoint[],
     path: string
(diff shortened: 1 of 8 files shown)


src/service.ts
```diff
@@ -207,36 +209,46 @@ export class DebugService implements IDebugger {
...
+    );
+    let kernelBreakpoints = reply.body.breakpoints.map(breakpoint => {
+      return {
+        ...breakpoint,
+        active: true,
+        source: { path: this.session.client.name }
+      };
+    });
+
+    // filter breakpoints with the same line number
```
jtpio: In theory an editor can display multiple breakpoints on the same line. But we can skip that case for now.

jtpio:
Still needs one more pass before being ready for review.

But https://github.com/jupyterlab/debugger/issues/130 can already base itself on it.

src/service.ts
JohanMabille: What use case do you have in mind for setting several breakpoints on the same line?

src/service.ts
```diff
@@ -207,36 +209,46 @@ export class DebugService implements IDebugger {
...
-    await this.session.sendRequest('setBreakpoints', {
+    return await this.session.sendRequest('setBreakpoints', {
       breakpoints: breakpoints,
       source: { path },
       sourceModified: false
     });
   };
 
-  updateBreakpoints = async () => {
+  updateBreakpoints = async (breakpoints: Breakpoints.IBreakpoint[]) => {
```
JohanMabille: I think that you need the same workaround as in `setBreakpoints` to avoid errors when switching to a new notebook for which the kernel is not ready.

src/service.ts
```diff
@@ -207,36 +209,46 @@ export class DebugService implements IDebugger {
...
-  getBreakpoints = (): DebugProtocol.SourceBreakpoint[] => {
-    return this._model.breakpointsModel.breakpoints.map(breakpoint => {
-      return {
-        line: breakpoint.line
-      };
-    });
-  };
-
   setBreakpoints = async (
     breakpoints: DebugProtocol.SourceBreakpoint[],
```
JohanMabille: Do we need to keep this method? We could move its implementation into `updateBreakpoints` since you are always supposed to send a `dumpCell` request before a `setBreakpoints` request.

JohanMabille: Actually we may need to keep it fo the case when the code changes in a cell: we may want to remove the breakpoints from the backend, in this case we already have the hash for the file name.

src/service.ts
jtpio: For example in the dev tools:

chrome-dev-tools-inline-breakpoints

But yeah as mentioned above it's fine to not do that and skip this feature for this debugger.

src/service.ts
jtpio: Yes we can keep it around for a while, but happy to move it to `updateBreakpoints` if it is not needed anymore.

src/service.ts
jtpio: Maybe we can keep the same workaround for both methods, since `updateBreakpoints` uses `setBreakpoints` at the moment.

jtpio: The call to `setBreakpoints` sometimes returns an empty response. This is happening when starting a fresh new notebook and its kernel session was not started before.

It looks like the debug messages are intertwined (most likely because of https://github.com/jupyterlab/debugger/issues/101)

jtpio:
@KsavinN about the issue you described earlier today with breakpoints showing up in the wrong cell.

I think this is because when 2 notebooks are open, the 2 `CellManager` are still there, and their signals still catch updates on the model:

https://github.com/jupyterlab/debugger/blob/dcf018a3f5b7a16804c3305c52484cbe2bb0f209/src/handlers/cell.ts#L28-L30

However the model is shared between all the cells, and at the moment it only contains the list of breakpoints for the active cell. So things get messed up easily.

Hopefully we'll get that sorted out with https://github.com/jupyterlab/debugger/issues/130

jtpio:
Ah ok just saw that you opened https://github.com/jupyterlab/debugger/issues/133.

src/service.ts
JohanMabille: This should be fixed with #130. No `setBreakpoints` request will be sent when starting a new notebook or switching to another one previously opened, only a restore session should be called, which will ask the backend for the breakpoints, update the model, and let the view update itself upon receiving signals.

PR #136, opened 2019-10-31
Use changed signal for models

### LINKED ISSUE #134: Streamline signals
We should be able to streamline all the signals and keep only one, that the views can use to be notified when the model has changed.

At the moment we use a couple of them, and it makes it difficult to track the different events. For example:

https://github.com/jupyterlab/debugger/blob/dcf018a3f5b7a16804c3305c52484cbe2bb0f209/src/handlers/cell.ts#L28-L34

Instead we can stick to one `changed` signal only:

- `currentLineChanged` is emitted when the stacktraces are set => this is a model change
- `linesCleared` is emitted on continued, and we set stack traces to `[]` => this is also a model change

jtpio:
Fixes #134 

All models should have a `changed` signal.

Some models still have additional signals such as `_currentVariableChanged` that have been kept for now (we might or might not need them later on).

This PR is building on top of #132 and should be rebased once #132 is merged into master.

For a cleaner diff of the change, see https://github.com/jtpio/debugger/pull/4

### DIFF
--- a/src/breakpoints/body.tsx
+++ b/src/breakpoints/body.tsx
@@ -35,10 +35,10 @@ const BreakpointsComponent = ({ model }: { model: Breakpoints.Model }) => {
       setBreakpoints(updates);
     };
 
-    model.breakpointsChanged.connect(updateBreakpoints);
+    model.changed.connect(updateBreakpoints);
 
     return () => {
-      model.breakpointsChanged.disconnect(updateBreakpoints);
+      model.changed.disconnect(updateBreakpoints);
     };
   });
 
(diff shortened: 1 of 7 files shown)


JohanMabille:
Now that #132 is merged, is this one ready for a review after a rebase? Or do you still have changes to push?

jtpio:
Yes it was left as a draft to avoid merging it by mistake. Now it's ready.

Also added a `TODO` to remove the `frames` from the service, that can be done separately:

https://github.com/jupyterlab/debugger/pull/136/files#diff-22ccf3ebb0cb6b300ee90a38b88edff8R292

PR #138, opened 2019-11-02
Fixed cell manager when debugger panel is closed and reopened

JohanMabille:
Currenlty, the debugger model is saved in the cell manager. The problem is that it is not correctly reset when the debugger panel is closed and the model is disposed.

Symetrically, when the debugger panel is opened again, the cell manager should be updated and the active cell should reset the current code value in the model.

Other handlers (notebook and consoles) should be fixed the same way since they keep a reference on the debugger model.

### DIFF
--- a/src/handlers/cell.ts
+++ b/src/handlers/cell.ts
@@ -21,30 +21,41 @@ const LINE_HIGHLIGHT_CLASS = 'jp-breakpoint-line-highlight';
 
 export class CellManager implements IDisposable {
   constructor(options: CellManager.IOptions) {
-    this._debuggerModel = options.debuggerModel;
     this._debuggerService = options.debuggerService;
-    this.breakpointsModel = options.breakpointsModel;
+    this.onModelChanged();
+    this._debuggerService.modelChanged.connect(() => this.onModelChanged());
     this.activeCell = options.activeCell;
     this.onActiveCellChanged();
-
-    this._debuggerModel.variablesModel.changed.connect(() => {
-      this.cleanupHighlight();
-      const firstFrame = this._debuggerModel.callstackModel.frames[0];
-      if (!firstFrame) {
-        return;
-      }
-      this.showCurrentLine(firstFrame.line);
-    });
-
-    this.breakpointsModel.changed.connect(async () => {
-      if (!this.activeCell || this.activeCell.isDisposed) {
-        return;
-      }
-      this.addBreakpointsToEditor(this.activeCell);
-    });
   }
 
   isDisposed: boolean;
+
+  private onModelChanged() {
+    this._debuggerModel = this._debuggerService.model;
+    if (this._debuggerModel) {
+      this.breakpointsModel = this._debuggerModel.breakpointsModel;
+
+      this._debuggerModel.variablesModel.changed.connect(() => {
+        this.cleanupHighlight();
+        const firstFrame = this._debuggerModel.callstackModel.frames[0];
+        if (!firstFrame) {
+          return;
+        }
+        this.showCurrentLine(firstFrame.line);
+      });
+
+      this.breakpointsModel.changed.connect(async () => {
+        if (!this.activeCell || this.activeCell.isDisposed) {
+          return;
+        }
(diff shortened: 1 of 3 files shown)


src/handlers/cell.ts
```diff
@@ -21,31 +21,42 @@ const LINE_HIGHLIGHT_CLASS = 'jp-breakpoint-line-highlight';
 
 export class CellManager implements IDisposable {
   constructor(options: CellManager.IOptions) {
-    this._debuggerModel = options.debuggerModel;
     this._debuggerService = options.debuggerService;
-    this.breakpointsModel = options.breakpointsModel;
+    this.onModelChanged();
+    this._debuggerService.modelChanged.connect(() => this.onModelChanged());
```
jtpio: We could also write it like to keep the usual style: 

```suggestion
    this._debuggerService.modelChanged.connect(this.onModelChanged, this);
```

src/handlers/cell.ts
```diff
@@ -21,31 +21,42 @@ const LINE_HIGHLIGHT_CLASS = 'jp-breakpoint-line-highlight';
...
+        if (!firstFrame) {
+          return;
+        }
+        this.showCurrentLine(firstFrame.line);
+      });
 
-    this.breakpointsModel.changed.connect(async () => {
-      if (!this.activeCell || this.activeCell.isDisposed) {
-        return;
+      this.breakpointsModel.changed.connect(async () => {
```
jtpio: I think these `connect()` also need a `disconnect()`, or the signal connections will start stacking up?

JohanMabille: I will clear the signal in the `dispose` method of the model.

afshin: You can do `Signal.clearData(this);` instead of disconnecting the individual signal, that will disconnect anybody who has connected as well.

src/breakpoints/index.ts
```diff
@@ -134,13 +134,25 @@ export namespace Breakpoints {
...
 
+    get isDisposed(): boolean {
+      return this._isDisposed;
+    }
+
+    dispose(): void {
+      if (!this._isDisposed) {
+        this._isDisposed = true;
+        Signal.clearData(this);
+      }
```
jtpio: Maybe we could write it like this to follow the current style (fast return)?

```typescript
    if (this.isDisposed) {
      return;
    }
    this._isDisposed = true;
    Signal.clearData(this);
```

src/handlers/cell.ts
```diff
@@ -21,31 +21,42 @@ const LINE_HIGHLIGHT_CLASS = 'jp-breakpoint-line-highlight';
...
 
-    this._debuggerModel.variablesModel.changed.connect(() => {
-      this.cleanupHighlight();
-      const firstFrame = this._debuggerModel.callstackModel.frames[0];
-      if (!firstFrame) {
-        return;
-      }
-      this.showCurrentLine(firstFrame.line);
-    });
+  isDisposed: boolean;
```
jtpio: Should we try to follow https://github.com/jupyterlab/debugger/issues/139 as of now and add a docstring? or later?

JohanMabille: let's do it later for the whole file in a "refactoring" pass.

src/breakpoints/index.ts
JohanMabille: Sure!

src/handlers/cell.ts
jtpio: :+1:

src/handlers/cell.ts
```diff
@@ -21,31 +21,42 @@ const LINE_HIGHLIGHT_CLASS = 'jp-breakpoint-line-highlight';
...
+
+  private onModelChanged() {
+    this._debuggerModel = this._debuggerService.model;
+    if (this._debuggerModel) {
```
jtpio: Maybe another "fast return" would be nice here, to make the whole block easier to read:

```typescript
if (!this._debuggerModel) {
    return;
}
```

PR #142, opened 2019-11-04
Add stepOut to the service and the command palette

jtpio:
Similar to `stepIn`.

image

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -137,6 +137,16 @@ export class DebugService implements IDebugger {
     }
   }
 
+  async stepOut(): Promise<void> {
+    try {
+      await this.session.sendRequest('stepOut', {
+        threadId: this.currentThread()
+      });
+    } catch (err) {
+      console.error('Error:', err.message);
+    }
+  }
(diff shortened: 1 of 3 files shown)


src/service.ts
```diff
@@ -137,6 +137,19 @@ export class DebugService implements IDebugger {
...
 
+  /**
+   * Makes the current thread step out a function / method if possible.
+   */
+  async stepOut(): Promise<void> {
+    try {
+      await this.session.sendRequest('stepOut', {
+        threadId: this.currentThread()
+      });
+    } catch (err) {
```
afshin: Do we want this `catch`? Or should the promise actually fail if the request fails?

jtpio: Eventually we'll want to remove the `try catch` and let the promise fail, so we can handle the error properly (showing an error message to the user for example).

jtpio: Thanks @afshin for the suggestion, opened https://github.com/jupyterlab/debugger/issues/147 to keep track of it.

Issue #139, opened 2019-11-03
Audit style conventions in TS classes


afshin:
We should follow the JupyterLab TS `class` conventions. These rules are how most JupyterLab and PhosphorJS classes are constructed.

Classes should be composed in this order:

1. `constructor` always appears first.
1. Public attributes (both `readonly` and those with getters/setters) appear next.
1. Public methods come next. The `public` keyword is unnecessary. All `public` methods should begin with a lowercase letter.
1. `protected` methods should appear next. All `protected` methods should begin with a lowercase letter.
1. `private` methods should follow. All `private` methods should begin with an `_` (underscore) so that as a user reads the code, they know they are looking at a `private` method/variable that cannot be inherited or used publicly.
1. `private` variables should appear last, at the bottom of a `class` definition.

If a `class` has static methods, the `static` keyword should not be used; *instead* the static methods should be defined in a separate `namespace` that has the same name as the class. In this way, `static` methods will not be conflated with `instance` methods.

cc: @KsavinN @jtpio @JohanMabille 

cf. [JupyterLab TypeScript Style Guide](https://github.com/jupyterlab/jupyterlab/wiki/TypeScript-Style-Guide)

afshin:
Here is an example of the pattern described above:

```typescript
export class Foo {
  constructor(options: Foo.IOptions = {}) {
    console.log('options', options);
  }

  readonly bar = true;

  get baz(): string {
    return this._baz;
  }
  set baz(baz: string) {
    this._baz = baz || 'empty';
  }

  qux(): void {
    console.log('Foo#qux() was called');
  }

  async quux(): Promise<void> {
    return;
  }

  protected quuz(): boolean {
    return Math.random() > 0.5;
  }

  protected async corge(): Promise<void> {
    this._grault();
    return this._garply();
  }

  private _grault(): void {
    console.log('Foo#_grault() was called');
  }

  private async _garply(): Promise<void> {
    return;
  }

  private _baz = '';
}

export namespace Foo {
  export interface IOptions {}

  export function create(): Foo {
    return new Foo();
  }
}
```

afshin:
Keep in mind that *all* methods (`public`, `protected` and `private`) should have doc strings.

All `public` and `protected` attributes should have doc strings as well.

jtpio:
Thanks @afshin for writing this down.

We should definitely follow this style every time we make a change. Maybe there are also some TSLint / ESLint rules that can help automated the process?

afshin:
[...]

PR #148, opened 2019-11-05
Notebook bp

JohanMabille:
This is an overview of what the breakpoint model could look like if we want it to track all the breakpoints of a notebook and restore them properly. I will split this PR into smaller ones so that it is easier to merge them (and also to take #139 into account), however review and comments on this one would be much appreciated. Remaining issues / questions for now:

- There is a bug in the breakpoint panel, when switching from a notebook with breakpoints to another notebook with breakpoints, sometimes a breakpoint appears twice or the breakpoints of both notebooks are merged. I guess this is due to the way react updates the component, but my limited skills in react do not allow me to understand what is going on

- I had to remove the Signal.clearData call when changing the client of a session, otherwise the events are not caught anymore when switching from a notebook to another but I am affraid that this would lead to stacking a lot of slots.

[Binder](https://mybinder.org/v2/gh/JohanMabille/debugger/notebook_bp?urlpath=lab/tree/examples/index.ipynb)

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -6,4 +6,4 @@ dependencies:
 - nodejs
 - notebook=6
 - ptvsd
-- xeus-python=0.6
+- xeus-python=0.6.4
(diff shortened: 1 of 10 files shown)


JohanMabille:
Superseded with #165

PR #152, opened 2019-11-05
Resolve setting breakpoints issue

### LINKED ISSUE #150: Breakpoint markers alignement
The hover markers and the breakpoint markers seem to be slightly misaligned:

image 

The fix would involve some CSS tweaking.

### LINKED ISSUE #151: Setting breakpoints on non-active cells
At the moment a cell must be active to be able to set breakpoints.

However the hover marker suggest that it should be possible to do so when a cell is not active:

breakpoints-non-active-cells

KsavinN:
resolve #151 
resolve #150 

chrome-capture (22)


jtpio:
Thanks @KsavinN.

Would it be possible to only keep the following commits for this PR (the ones related to this change)?

- https://github.com/jupyterlab/debugger/pull/152/commits/177ab586a1cfbf79457b5c704c01c01c12a8fe74
- https://github.com/jupyterlab/debugger/pull/152/commits/7602425ca0b104fb936c939ed6593e6f8f029fb1

The other commits seem to be coming from other PRs and add a bit of noise to the git history.

PR #153, opened 2019-11-05
Restore debugger state

JohanMabille:
This also includes a first pass on `IDebugger`, `IDebugger.ISession`, `DebugService` and `DebugSession` according to #139

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -3,8 +3,6 @@
 
 import { IClientSession } from '@jupyterlab/apputils';
 
-import { CodeEditor } from '@jupyterlab/codeeditor';
-
 import { KernelMessage } from '@jupyterlab/services';
 
 import { PromiseDelegate } from '@phosphor/coreutils';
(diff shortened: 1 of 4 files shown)


src/session.ts
```diff
@@ -24,29 +25,39 @@ export class DebugSession implements IDebugger.ISession {
...
+   * @param client - The new API client session.
+   */
   set client(client: IClientSession | null) {
     if (this._client === client) {
       return;
     }
 
     if (this._client) {
-      Signal.clearData(this._client);
+      this._client.iopubMessage.disconnect(this._handleEvent, this);
```
jtpio: And `Signal.clearData(this._client.iopubMessage)` would not be suitable because it would clear too much?

JohanMabille: Yes, we don't want to remove other signals, only the one of the debugger session.

JohanMabille: Indeed, we don't want to remove any other slot, only the one of this session.

jtpio: Good, thanks for the confirmation :+1:

jtpio:
Thanks!

Merging so we can iterate on the restore flow.

PR #156, opened 2019-11-05
resolve hover issue and aligment of gutter

### LINKED ISSUE #150: Breakpoint markers alignement

### LINKED ISSUE #151: Setting breakpoints on non-active cells

KsavinN:
resolve #151
resolve #150
image

### DIFF
--- a/style/breakpoints.css
+++ b/style/breakpoints.css
@@ -9,7 +9,7 @@
 
 .jp-breakpoint-marker {
   position: absolute;
-  left: -35px;
+  left: -34px;
   top: -1px;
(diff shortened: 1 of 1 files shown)


jtpio:
Looks like GitHub had an incident and this PR is in a broken state:

image

jtpio:
The commits show up in the history:

image

Issue #128, opened 2019-10-30
Add support for disabled breakpoint


jtpio:
Support for disabling breakpoints would require:

- displaying disabled breakpoints in grey (or similar color)
- handling the change on the `active` state of the `BreakpointComponent`
- trigger a new call to `updateBreakpoints` filtering the ones that are not active (disabled)

![image](https://user-images.githubusercontent.com/591645/67845815-6c79bc80-fb00-11e9-9d29-9613fe1469d3.png)

jtpio:
We could think of a "filter" that is easy to set and unset on the model, which could also be reused for other functionalities such as the `stop` button (see https://github.com/jupyterlab/debugger/pull/157).

jtpio:
This would require a change in the backend (where the debug state is), so the kernel is able to store the list of deactivated breakpoints.

jtpio:
For reference, UI elements for deactivating breakpoints were removed in https://github.com/jupyterlab/debugger/pull/228.

PR #157, opened 2019-11-06
Add the stop button to the callstack toolbar

### LINKED ISSUE #155: Stop button
If we still want to keep the `stop` button, we can implement what is mentioned in https://github.com/jupyterlab/debugger/issues/81#issuecomment-547799815:

> The stop button should continue the execution until the entire cell has been executed. This can be done thanks to an additional method in the service that automagically sends continue when the code hits a breakpoint.

### TODO

- Add a `stop` method to the `DebugService`
- Add a `stop` command that uses this method
- Add a `stop` button in the callstack toolbar
- Create tests for the `DebugService` (`tests/service.spec.ts`)

jtpio:
Fixes #155 

There doesn't seem to be any straightforward way to continue the execution of a cell while ignoring all the breakpoints.

Altering the debug session state (removing and re-adding the breakpoints) could work, but could lead to edge cases such as the page being reloaded and losing all the breakpoints.

Another approach would be to support `disabled` breakpoints (#128):
- Click on `stop`
- Disable all breakpoints
- Wait for `execution_reply`
- Enable all breakpoints (or only the ones that were enabled before the `stop`)

### TODO

- [x] Add the `stop` button
- [x] Clear the UI on `stop`
- [x] Send breakpoints after restarting the debugger

stop-button-3

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -299,8 +299,7 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
         return service.isThreadStopped();
       },
       execute: async () => {
-        await service.stop();
-        await service.start();
+        await service.restart();
         commands.notifyCommandChanged();
       }
     });
--- a/src/index.ts
+++ b/src/index.ts
@@ -65,8 +65,6 @@ export namespace CommandIDs {
 
   export const debugFile = 'debugger:debug-file';
 
-  export const debugNotebook = 'debugger:debug-notebook';
-
   export const mount = 'debugger:mount';
 
   export const changeMode = 'debugger:change-mode';
@@ -411,13 +409,17 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
 
     if (palette) {
       const category = 'Debugger';
-      palette.addItem({ command: CommandIDs.changeMode, category });
-      palette.addItem({ command: CommandIDs.create, category });
-      palette.addItem({ command: CommandIDs.debugContinue, category });
-      palette.addItem({ command: CommandIDs.next, category });
-      palette.addItem({ command: CommandIDs.stepIn, category });
-      palette.addItem({ command: CommandIDs.stepOut, category });
-      palette.addItem({ command: CommandIDs.debugNotebook, category });
+      [
+        CommandIDs.changeMode,
+        CommandIDs.create,
+        CommandIDs.debugContinue,
+        CommandIDs.terminate,
+        CommandIDs.next,
+        CommandIDs.stepIn,
+        CommandIDs.stepOut
+      ].forEach(command => {
+        palette.addItem({ command, category });
+      });
     }
 
     if (restorer) {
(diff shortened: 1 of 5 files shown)


jtpio:
This is what it looks like at the moment:

stop-button

src/service.ts
```diff
@@ -394,7 +396,7 @@ export class DebugService implements IDebugger {
   private frames: Frame[] = [];
 
   // TODO: move this in model
-  private _threadStopped = new Set();
+  private _stoppedThreads = new Set();
```
jtpio: `_threadStopped` made it look like a `Signal`.

jtpio:
Also added to the palette for consistency:

image

src/index.ts
```diff
@@ -291,6 +291,20 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
...
+    commands.addCommand(CommandIDs.terminate, {
+      label: 'Terminate',
+      caption: 'Terminate',
+      iconClass: 'jp-MaterialIcon jp-StopIcon',
+      isEnabled: () => {
+        return service.isThreadStopped();
+      },
+      execute: async () => {
+        await service.stop();
+        await service.start();
```
jtpio: This could actually be moved to a `service.terminate()` method, which will also contain the logic below (cleaning `clearModel`, `_stoppedThreads` and so on)

jtpio: Or better: `service.restart()` as this is what it is doing at the moment.

JohanMabille: Indeed that is a better name. I think we should wait for the execution to complete before restarting (or at least before sending the breakpoints back). Otherwise the code may stop again if a computation is long enough so that the sevice has the time to set the breakpoints again.

jtpio: Sending the breakpoints after restart, but without explicitly waiting for the `execute_reply`:
 
stop-button-2

src/index.ts
```diff
@@ -291,6 +291,19 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
       }
     });
 
+    commands.addCommand(CommandIDs.terminate, {
+      label: 'Terminate',
```
jtpio: We keep the name `Terminate` as it feels more natural than `Restart`.

jtpio:
This is the current logic of the `restart` as of now:

```typescript
async restart(): Promise<void> {
	const breakpoints = this.model.breakpointsModel.breakpoints;
	await this.stop();
	this.clearModel();
	this._stoppedThreads.clear();
	await this.start();
	await this.updateBreakpoints(breakpoints);
}
```

This means that the breakpoints from the other cells will not be restored. But they should be after #130.

jtpio:
This change is self-contained enough (all behind the `terminate` command), so marking it as ready for review now.

To try it on Binder: https://mybinder.org/v2/gh/jtpio/debugger/stop-button?urlpath=/lab/tree/examples/index.ipynb

PR #160, opened 2019-11-07
Remove onNewRenderLine

### LINKED ISSUE #159: Error when adding new lines to cells with breakpoints
Following these steps:

- Add a breakpoint to a cell
- Edit the cell by adding a new line

Triggers this error:

```
Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node
```

new-lines-cell-bps

jtpio:
Fixes #159.

Not exactly sure why we had this in the first place. But the `onNewRenderLine` was modifying the breakpoints model directly and triggering CodeMirror issues (#159).

Instead we'll want to send the `dumpCell` and `setBreakpoints` requests via the service when a user modifies the content of a cell.

remove-renderline

### DIFF
--- a/src/breakpoints/index.ts
+++ b/src/breakpoints/index.ts
@@ -118,22 +118,6 @@ export namespace Breakpoints {
       this.breakpoints = this._state[newType];
     }
 
-    changeLines(lines: number[]) {
-      if (!lines && this.breakpoints.length === 0) {
-        return;
-      }
-      if (lines.length === 0) {
-        this.breakpoints = [];
-      } else {
-        const breakpoint = { ...this.breakpoints[0] };
-        let breakpoints: Breakpoints.IBreakpoint[] = [];
-        lines.forEach(line => {
-          breakpoints.push({ ...breakpoint, line });
-        });
-        this.breakpoints = [...breakpoints];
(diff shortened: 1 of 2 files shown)


jtpio:
cc @KsavinN if you want to have a look

KsavinN:
There is good news and bad:

Realy great is that fix #159 

but its seems we need found solutions for this:
example in vs-code:
Peek 2019-11-07 15-43

and now in this PR:
bad

jtpio:
Yes that's right.

But since we need to sync the content of the cell when it is modified on the frontend, we'll also have to update the breakpoints. The flow would be something like this:

- Add a signal connection to `activeCell.model.contentChanged`.
- When a user changes the content of the cell, the signal will be emitted.
- Send `dumpCell` and `setBreakpoints`, probably behind an `ActivityMonitor` or `Debouncer` to avoid sending to many requests.
- Since the service is responsible for updating the breakpoints, the view will refresh itself when a response has been received from the kernel and the model changed.

jtpio:
Opened https://github.com/jupyterlab/debugger/issues/162 so we can track that in a separate issue.

PR #164, opened 2019-11-08
Update breakpoints on cell content changed

### LINKED ISSUE #162: Update cell and breakpoints on content changed
For the debugger to work, we need keep the content of the cell in sync with the kernel. This is because of the hashing mechanism detailed in https://github.com/jupyterlab/debugger/issues/49 and https://github.com/jupyterlab/debugger/issues/52.

That means that every time a user modifies the content of the cell, we need to send the `dumpCell` and `setBreakpoints` requests, for the debugger to be able to break at the next `execute_request`.

The flow would be something similar to the following:

- Add a signal connection to `activeCell.model.contentChanged`.
- When a user changes the content of the cell, the signal is emitted.
- Send `dumpCell` and `setBreakpoints` via the `DebugService`, probably behind an `ActivityMonitor` or `Debouncer` to avoid firing to many requests.
- Since the service is responsible for updating the breakpoints, the breakpoints view will refresh itself when a response is received from the kernel and the model changed.

This was also raised in https://github.com/jupyterlab/debugger/pull/160. Before the breakpoints model would be directly modified when the number of lines of the cell changes (and the service bypassed).

jtpio:
Fixes #162.

The idea is to send the content of the cell and the list of breakpoints to the kernel, so that the next `execute_request` will be able to break using the up-to-date cell code.

cell-activity

### TODO

- [x] Update breakpoints on cell content changed
- [x] Use an `ActivityMonitor` (or `Debouncer`) to throttle the updates

### DIFF
--- a/src/handlers/cell.ts
+++ b/src/handlers/cell.ts
@@ -139,11 +139,40 @@ export class CellManager implements IDisposable {
       this._debuggerService.session
     ) {
       if (this.previousCell && !this.previousCell.isDisposed) {
+        this.previousCell.model.contentChanged.disconnect(
+          this.sendEditorBreakpoints,
+          this
+        );
         this.removeListener(this.previousCell);
       }
+      this.activeCell.model.contentChanged.connect(
+        this.sendEditorBreakpoints,
+        this
+      );
       this.previousCell = this.activeCell;
       this.setEditor(this.activeCell);
     }
+  }
+
+  protected sendEditorBreakpoints() {
+    // TODO: put behind a Debouncer / ActivityMonitor
+    const cell = this.activeCell;
+    if (!cell || !cell.editor) {
+      return;
+    }
+
+    const breakpoints = this.getBreakpointsFromEditor(cell).map(lineInfo => {
+      return Private.createBreakpoint(
+        this._debuggerService.session.client.name,
+        this.getEditorId(),
+        lineInfo.line + 1
+      );
+    });
+
+    void this._debuggerService.updateBreakpoints(
+      cell.editor.model.value.text,
+      breakpoints
+    );
   }
 
   protected setEditor(cell: CodeCell) {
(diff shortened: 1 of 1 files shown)


jtpio:
One downside is that the cell can be executed before the ActivityMonitor has time to kick-in (during the timeout).

jtpio:
We could also update the breakpoints when the notebook is saved.

jtpio:
This PR should now be in a state that can be merged.

However we may want to reconsider how to update the breakpoints and sync the content of the cells.

Relying on the activity monitor leaves the door open for some edge cases, such as executing the code before the cell is dumped to the kernel.

jtpio:
This is the what can happen when modifying a cell and executing before there is time to sync it:

cell-activity-sync

src/handlers/cell.ts
```diff
@@ -216,6 +256,18 @@ export class CellManager implements IDisposable {
     });
   }
 
+  private getBreakpointsFromEditor(cell: CodeCell): ILineInfo[] {
```
jtpio: This method was removed in https://github.com/jupyterlab/debugger/pull/165, but we still need to retrieve the breakpoints from the editor when the content of the cell changes.

jtpio: cc @JohanMabille 

It looks like we need to keep this method, or we might end up in situations such as the one described here: https://github.com/jupyterlab/debugger/issues/172#issue-521721030

Also it feels more natural to have the breakpoints stay where they are when editing a cell.

What do you think?

jtpio:
Rebased.

src/handlers/cell.ts
JohanMabille: Indeed this methods should be added back. I removed it when I refactored the logic for setting breakpoints because it was not called anymore, but it should definitely be reintroduced.

jtpio: ok good :+1:

PR #165, opened 2019-11-08
The model now keeps all the breakpoints of a notebook

JohanMabille:
This PR breaks two features that need further refactoring to be restored:
- breakpoints activation from the breakpoint panel
- ~~restart does not send the breakpoints back to the kernel~~

EDIT: last feature has been restored.

### DIFF
--- a/src/breakpoints/body.tsx
+++ b/src/breakpoints/body.tsx
@@ -2,7 +2,6 @@
 // Distributed under the terms of the Modified BSD License.
 
 import { ReactWidget } from '@jupyterlab/apputils';
-import { ArrayExt } from '@phosphor/algorithm';
 import { ISignal } from '@phosphor/signaling';
 import React, { useEffect, useState } from 'react';
 import { Breakpoints } from '.';
@@ -22,26 +21,52 @@ export class Body extends ReactWidget {
 }
 
 const BreakpointsComponent = ({ model }: { model: Breakpoints.Model }) => {
-  const [breakpoints, setBreakpoints] = useState(model.breakpoints);
+  const [breakpoints, setBreakpoints] = useState(
+    Array.from(model.breakpoints.entries())
+  );
 
   useEffect(() => {
     const updateBreakpoints = (
       _: Breakpoints.Model,
       updates: Breakpoints.IBreakpoint[]
     ) => {
-      if (ArrayExt.shallowEqual(breakpoints, updates)) {
-        return;
-      }
-      setBreakpoints(updates);
+      setBreakpoints(Array.from(model.breakpoints.entries()));
+    };
+
+    const restoreBreakpoints = (_: Breakpoints.Model) => {
+      setBreakpoints(Array.from(model.breakpoints.entries()));
     };
 
     model.changed.connect(updateBreakpoints);
+    model.restored.connect(restoreBreakpoints);
 
     return () => {
       model.changed.disconnect(updateBreakpoints);
+      model.restored.disconnect(restoreBreakpoints);
     };
   });
 
+  return (
+    <div>
+      {breakpoints.map(entry => (
+        // Array.from(breakpoints.entries()).map((entry) => (
+        <BreakpointCellComponent
+          key={entry[0]}
+          breakpoints={entry[1]}
+          model={model}
+        />
+      ))}
+    </div>
+  );
+};
+
+const BreakpointCellComponent = ({
+  breakpoints,
+  model
+}: {
+  breakpoints: Breakpoints.IBreakpoint[];
+  model: Breakpoints.Model;
+}) => {
   return (
     <div>
       {breakpoints
@@ -50,7 +75,7 @@ const BreakpointsComponent = ({ model }: { model: Breakpoints.Model }) => {
         })
         .map((breakpoint: Breakpoints.IBreakpoint) => (
           <BreakpointComponent
-            key={breakpoint.line}
+            key={breakpoint.source.path + breakpoint.line}
             breakpoint={breakpoint}
             breakpointChanged={model.breakpointChanged}
           />
--- a/src/breakpoints/index.ts
+++ b/src/breakpoints/index.ts
@@ -5,7 +5,6 @@ import { Toolbar, ToolbarButton } from '@jupyterlab/apputils';
 import { IDisposable } from '@phosphor/disposable';
 import { Signal } from '@phosphor/signaling';
 import { Panel, PanelLayout, Widget } from '@phosphor/widgets';
-import { murmur2 } from 'murmurhash-js';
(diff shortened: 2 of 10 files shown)


jtpio:
> This PR breaks two features that need further refactoring to be restored:
> 
>     * breakpoints activation from the breakpoint panel
> 
>     * restart does not send the breakpoints back to the kernel

I think it's fine to do that after. The breakpoints activation requires some additional logic anyway by keeping a state of the currently deactivated breakpoints (#128).

src/breakpoints/body.tsx
```diff
@@ -22,26 +21,52 @@ export class Body extends ReactWidget {
...
     ) => {
-      if (ArrayExt.shallowEqual(breakpoints, updates)) {
-        return;
-      }
-      setBreakpoints(updates);
+      setBreakpoints(Array.from(model.breakpoints.entries()));
+    };
+
+    const restoreBreakpoints = (_: Breakpoints.Model) => {
+      setBreakpoints(Array.from(model.breakpoints.entries()));
```
jtpio: This looks like the `restored` and `changed` update logic could be merged into one?

src/breakpoints/index.ts
```diff
@@ -130,13 +111,32 @@ export namespace Breakpoints {
...
+      this._breakpoints.set(id, breakpoints);
+      this.changed.emit(breakpoints);
+    }
+
+    getBreakpoints(id: string): IBreakpoint[] {
+      return this._breakpoints.get(id) || [];
+    }
+
+    restoreBreakpoints(breakpoints: Map<string, IBreakpoint[]>) {
+      this._breakpoints = breakpoints;
```
jtpio: Is there a reason for not using the `breakpoints` setter that will emit the `changed` signal?

jtpio: Never mind that comment, the setter is removed in this change.

jtpio:
Testing the change locally (after a `jlpm run clean` and updating `xeus-python`), I noticed the following behavior where the breakpoints are not shown in the gutters. Also they can be added but not removed.

breakpoints-model

src/breakpoints/index.ts
jtpio: So it looks like the main difference between `setBreakpoints` and `restoreBreakpoints` is that `restoreBreakpoints` sets all breakpoints at once, while `setBreakpoints` will do it id by id.

Would it make sense to also use the `changed` signal when restoring the breakpoints, or do we expect the restore logic to be different later and require an additional `restored` signal?

jtpio:
> Testing the change locally (after a `jlpm run clean` and updating `xeus-python`), I noticed the following behavior where the breakpoints are not shown in the gutters.

This seems to be because the `setBreakpoints` method is called with the code of the cell as the `id`, while the Map stores the hash of the code.

src/service.ts
```diff
@@ -284,10 +327,25 @@ export class DebugService implements IDebugger {
       (breakpoint, i, arr) =>
         arr.findIndex(el => el.line === breakpoint.line) === i
     );
-    this._model.breakpointsModel.breakpoints = kernelBreakpoints;
+    this._model.breakpointsModel.setBreakpoints(code, kernelBreakpoints);
```
jtpio: Should that be `dumpedCell.sourcePath` instead?

```suggestion
    this._model.breakpointsModel.setBreakpoints(dumpedCell.sourcePath, kernelBreakpoints);
```

jtpio: Or calculating the cell id with `this._debuggerService.getCellId(cell.model.value.text)`

src/breakpoints/index.ts
```diff
@@ -130,13 +111,32 @@ export namespace Breakpoints {
...
+    }
+
+    /*private printMap() {
+      this._breakpoints.forEach((value, key, map) => {
+        console.log(key);
+        value.forEach((bp) => console.log(bp.line));
+      });
+    }*/
+
+    private _breakpoints = new Map<string, IBreakpoint[]>();
```
jtpio: Should we make an `interface` out of it?

src/breakpoints/body.tsx
JohanMabille: This depends whether we want to keep the current logic, where we ask for all the breakpoints each time there is an update instead of emitting the breakpoints that changed in the signal.
In the first case, we can change the signal signature and merge the `restored` and `update` functions.

src/breakpoints/index.ts
JohanMabille: I'm not sure yet (see my reply to your comment about the `restore`/`update` functions in `body.tsx`

src/service.ts
JohanMabille: :sweat_smile:

src/breakpoints/index.ts
JohanMabille: Maybe a type alias would be enough?

jtpio: Sounds good :+1:

src/breakpoints/body.tsx
jtpio: How about we leave it like this and open a new issue, so we can have a look again later? (if needed)

src/breakpoints/index.ts
jtpio: :+1:

jtpio:
> EDIT: last feature has been restored.

Want to add a test for it in this PR? (https://github.com/jupyterlab/debugger/issues/166)
Or as a separate change?

JohanMabille:
let's do it in this PR.

JohanMabille:
Ok actually let's do it in another PR, there are also refactoring to do in the tests, most of the tests done in `session.spec.ts` should actuvally be done in `service.spec.ts`.

jtpio:
OK.

Yes for now the tests have been for the debug session to serve as reference.

src/breakpoints/index.ts
```diff
@@ -130,13 +111,32 @@ export namespace Breakpoints {
...
+      this._breakpoints = breakpoints;
+      this._restored.emit();
+    }
+
+    /*private printMap() {
+      this._breakpoints.forEach((value, key, map) => {
+        console.log(key);
+        value.forEach((bp) => console.log(bp.line));
+      });
+    }*/
```
jtpio: Used for debugging?

jtpio:
It looks like clearing all the breakpoints at once will perform the clear only the cell gets the focus:

clear-all-breakpoints

We can log that in a separate issue for now.

JohanMabille:
Yes, because cells that do not have the focus are not connected to the signal. We need to maintain a mapping id <-> cells to be able to clean all the gutters, but I prefer to do that in a dedicated PR (there are already a lot of potential conflicting changes in this one).

jtpio:
Sounds good :+1:

Issue #169, opened 2019-11-12
Open files in the tab panel editors widget


jtpio:
At the moment, the read-only tab panel editors in the main area widget display placeholder data:

![image](https://user-images.githubusercontent.com/591645/68668687-10c02200-0549-11ea-8c91-caef4aa04ad3.png)

Instead, we should implement the idea mentioned in the [design](./DESIGN.md) document:

> - In expanded mode, the debugger will contain a tab panel for text editors that are launched when a user steps into code that has been stopped via breakpoints.
> - If the user adds a breakpoint and steps into code from the sidebar, then it should automatically switch to expanded mode to accommodate displaying code files.
> - Code files in the debugger will always be read-only. They should support adding breakpoints.

jtpio:
There are 2 points worth keeping in mind:

- Temporary files are store in folders such as `/tmp/xpython_12345/`, which are not under the JupyterLab root 
- Remote kernels: kernels are not local to the Jupyter server (this scenario could be ignored for now)

jtpio:
Actually we might be able to use the `loadedSources` request from the DAP:

https://microsoft.github.io/debug-adapter-protocol/specification#Requests_LoadedSources

jtpio:
One behavior could be:

- `stopped` event received
- send `stackTraces` request
- from the `stackFrames` response, get the `source.path`
- check if `source.path` matches any of the existing cells / editors (with a lookup or looping through all the cells)
- if it does, focus on the cell and highlight the line there
- otherwise send the `source` request
- show the content of the response in a read-only editor, and highlight the line in the editor

jtpio:
https://github.com/jupyterlab/debugger/pull/170 adds tests for the `loadedSources` and `source` requests.

At the moment they don't seem to be recognized though. Investigating.

jtpio:
As discussed with @JohanMabille, we could as a workaround implement the `source` request in `xeus-python` directly and still comply to the DAP.

jtpio:
`source` was added to the latest `xeus-python` (0.6.6): https://github.com/QuantStack/xeus-python/pull/168

This should make it possible to open the files in the read-only editors.

PR #170, opened 2019-11-12
Add tests for source and loadedSources

jtpio:
This change adds two tests for the `DebugSession`.

This is in preparation for #169, and to test whether it is possible to retrieve the content for a `DebugProtocol.Source`.

At the moment the tests fail for both the `loadedSources` and `sources` request:

```
D00000.415: (while handling {'command': 'loadedSources', 'seq': 5, 'type': 'request'})
            IDE <-- {
                "type": "response",
                "seq": 9,
                "request_seq": 5,
                "success": false,
                "command": "loadedSources",
                "message": "Unknown command",
                "body": {}
            }
```

```
D00000.599: (requested while handling {'command': 'source', 'seq': 7, 'type': 'request'})
            PYD --> {
                "type": "response",
                "request_seq": 1000000005,
                "success": false,
                "command": "source",
                "body": {
                    "content": ""
                },
                "seq": 22,
                "message": "Source unavailable",
                "pydevd_cmd_id": 502
            }
```

### DIFF
--- a/tests/src/session.spec.ts
+++ b/tests/src/session.spec.ts
@@ -262,4 +262,31 @@ describe('protocol', () => {
       expect(j.value).to.equal('4');
     });
   });
+
+  describe('#loadedSources', () => {
+    it('should retrieve the list of loaded sources', async () => {
+      const loadedSourcesReply = await debugSession.sendRequest(
+        'loadedSources',
+        {}
+      );
+      const sources = loadedSourcesReply.body.sources;
+      expect(sources).to.exist;
+      expect(sources.length).to.be.above(0);
+    });
+  });
+
+  describe('#source', () => {
+    it('should retrieve the source of the dumped code cell', async () => {
+      const stackFramesReply = await debugSession.sendRequest('stackTrace', {
+        threadId
+      });
+      const frame = stackFramesReply.body.stackFrames[0];
+      const source = frame.source;
+      const reply = await debugSession.sendRequest('source', {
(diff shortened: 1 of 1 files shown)


jtpio:
Will check what happens in VS Code and whether or not the `source` request is sent.

jtpio:
Testing with:

`test.py`

```python
from mymodule import add

i = 0
i += 1
i += 2
i

add(i, 2)
```

`mymodule.py`

```python
def add(a, b):
    print(a)
    print(b)
    return a + b
```

Stepping into the `add` function triggers the `module` event, with the path to `mymodule.py`. But there is no `source` request, most likely because the file is local and VS code is able to open it.

jtpio:
Rebased so we can use the latest `xeus-python` with support for `source`.

tests/src/session.spec.ts
```diff
@@ -262,4 +262,28 @@ describe('protocol', () => {
...
     });
   });
+
+  describe('#loadedSources', () => {
+    it('should *not* retrieve the list of loaded sources', async () => {
+      // `loadedSources` is not supported at the moment "unknown command"
+      const reply = await debugSession.sendRequest('loadedSources', {});
+      expect(reply.success).to.be.false;
+    });
+  });
```
jtpio: Leaving it in the tests so we know `loadedSources` is not supported.

Also fine to remove it if needed.

JohanMabille: I think it's a good thing to keep it.

Issue #144, opened 2019-11-04
Behavior of next / stepOver


jtpio:
We should check what it happening when the `stepOver` (`next`) command is being used.

In the screencast below we would expect to advance to the line 2, then 3 and then 4 when doing several calls to `stepOver`.

![step-over-behavior](https://user-images.githubusercontent.com/591645/68118641-c6afbe80-ff00-11e9-99f9-876ccd481105.gif)

jtpio:
One difference compared to VS code is that we always seem to have one more scope on line 2. Which would explain why the `next` command will add that "extra jump": 

![step-over-behavior-2](https://user-images.githubusercontent.com/591645/68378501-7fb60900-014c-11ea-9aa0-c42519ee9ce2.gif)

jtpio:
As a side-effect, we can continue breaking and stepping even after removing the breakpoints:

![step-over-behavior-3](https://user-images.githubusercontent.com/591645/68378830-1387d500-014d-11ea-9ceb-3594540f6cf6.gif)

jtpio:
Which is most likely related to how the code is executed in `xeus-python`:

- the `<string>` source coming from: https://github.com/QuantStack/xeus-python/blob/507786ce6e726ba3ea518f99fce5636e2e46d329/src/xinterpreter.cpp#L126
- the execution: https://github.com/QuantStack/xeus-python/blob/507786ce6e726ba3ea518f99fce5636e2e46d329/src/xinterpreter.cpp#L150-L156

![image](https://user-images.githubusercontent.com/591645/68379510-42eb1180-014e-11ea-9483-1f4900f0b2a6.png)

JohanMabille:
I wonder if this behavior does not come from the way we initialize ptvsd. Indeed, I can observe the same behavior even if the last statement is not an expression to evaluate. I've also played a bit with the backend, the following change doe snot fix the issue:

```cpp
    std::string filename = get_cell_tmp_file(code);
    register_filename_mapping(filename, execution_count);
    
    // Parse code to AST
    py::object code_ast = ast.attr("parse")(code, filename, "exec");
    py::list expressions = code_ast.attr("body");
```

jtpio:
> I wonder if this behavior does not come from the way we initialize ptvsd

That is also one possibility yes.

JohanMabille:
This issue is definitely due to the backend implementation. It seems that ptvsd doe snot support interactive mode yet (see https://github.com/microsoft/ptvsd/issues/201).

[...]

PR #182, opened 2019-11-14
Change line highlighting when switching callstack

### LINKED ISSUE #173: Switch callstack
We already have support for updating the variables when switching callstack:

switch-callstacks

The next step would be to show the associated code and line highlighting when doing so, similar to this:

vscode-switch-callstack

jtpio:
Fixes #173.

Leaving the "jump to code" aspect of #173 out of this scope for now (requires adding support for lookup using the hashes first).

switch-callstacks2

### TODO

- [x] Show line highlight for the current selected frame
- [x] Remove `frames` from the `DebugService` class

### DIFF
--- a/src/callstack/index.ts
+++ b/src/callstack/index.ts
@@ -96,6 +96,7 @@ export namespace Callstack {
     set frames(newFrames: IFrame[]) {
       this._state = newFrames;
       this._framesChanged.emit(newFrames);
+      this.frame = newFrames[0];
     }
 
     get frames(): IFrame[] {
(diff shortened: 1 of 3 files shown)


jtpio:
The last commit removes the `frames` from the `DebugService`.

This means that scopes are being fetched every time we switch between frames, instead of fetching them all at once:

switch-callstacks3

jtpio:
Note: the frame on line 2 is related to https://github.com/jupyterlab/debugger/issues/144

KsavinN:
check locally. It works well

PR #190, opened 2019-11-14
Resolve-switching between notebooks removes the breakpoints from the UI

### LINKED ISSUE #185: Dark theme icons
Some icons are not visible when using the dark theme:

image

image

### LINKED ISSUE #186: Reopening a notebook with an active session throws an error
To reproduce:

1. Open a notebook with `xpython` as a kernel
2. Add a breakpoint, execute, continue
3. Close the notebook
4. Reopen the notebook
5. Repeat step 2

open-notebook-active-kernel-null-editor

### LINKED ISSUE #187: Switching between notebooks removes the breakpoints from the UI
This is more of a visual glitch than a bug, as the breakpoints are correctly restored when focusing back on the cell. But it can be confusing to see them disappearing and re-appearing:

breakpoints-hidden-switching-notebooks

### LINKED ISSUE #197: First start
Starting the debugger fresh from a new workspace doesn't seem to be picked up by the handlers:

debugger-first-start

It requires switching activities and bringing the focus back on the notebook.

### LINKED ISSUE #204: No clear handlers after dispose debugger
disposeDebugger

probably related with #197

KsavinN:
Resolve #187 
Resolve #186 
Resolve #197 
Resolve #204 
Resolve #185  

**edit**: format message to close all issues at once


jtpio:
Thanks @KsavinN.

That's fixing a lot of different issues, so maybe we could have split the changes into multiple PRs?

Is the state still work in progress?

KsavinN:
> Thanks @KsavinN.
> 
> That's fixing a lot of different issues, so maybe we could have split the changes into multiple PRs?
> 
> Is the state still work in progress?

I think it's done. Yeah, we can split that or rename PR.

src/handlers/notebook.ts
```diff
@@ -53,21 +59,26 @@ export class NotebookHandler implements IDisposable {
       return;
     }
     this.isDisposed = true;
+    this.cleanAllCells();
     this.cellManager.dispose();
     Signal.clearData(this);
   }
 
+  protected cleanAllCells() {
```
jtpio: There is also some cleanup happening a bit below in the `onCurrentFrameChanged` method:

```typescript
cells.forEach(cell => CellManager.cleanupHighlight(cell));
```

Should we try to group that together?

KsavinN: I make new exported function now for setOff line numbers.

src/handlers/cell.ts
```diff
@@ -302,6 +314,16 @@ export namespace CellManager {
...
   }
+
+  /**
+   * Remove line numbers and all gutters from cell.
+   * @param cell The cell to cleanup.
+   */
+
+  export function setOffOptions(cell: Cell) {
+    const editor = cell.editor as CodeMirrorEditor;
+    editor.setOption('lineNumbers', false);
```
jtpio: I haven't tested locally yet, but does that mean the line numbers will keep switching between the on and off states?

This was changed in https://github.com/jupyterlab/debugger/pull/114 to fix https://github.com/jupyterlab/debugger/issues/106.

Or a screencast could be helpful here to see what it looks like.

KsavinN: This setOffOptions is executed only when Debbugger widget is closed.
example

jtpio: But line numbers could already be enabled before using the debugger:

image

Wondering how that would play together.

KsavinN: You mean that user set by default that or just set by default by us ?

jtpio: Yes by a user. It's also possible to set it via the settings editor:

image

KsavinN: ok. i will just implement clearAllGutters

src/index.ts
```diff
@@ -217,7 +227,7 @@ const notebooks: JupyterFrontEndPlugin<void> = {
  */
 const main: JupyterFrontEndPlugin<IDebugger> = {
   id: '@jupyterlab/debugger:main',
-  optional: [ILayoutRestorer, ICommandPalette],
+  optional: [ILayoutRestorer, ICommandPalette, ILabShell],
```
jtpio: Should we make `ILabShell` part of the `requires`? (similar to the other plugins)

KsavinN: Good point, I resolved that.

jtpio: Thanks!

src/handlers/cell.ts
```diff
@@ -145,8 +133,11 @@ export class CellManager implements IDisposable {
         this.sendEditorBreakpoints();
       }, this);
 
+      requestAnimationFrame(() => {
```
jtpio: Why do we need the `requestAnimationFrame`?

src/handlers/cell.ts
```diff
@@ -30,10 +30,10 @@ const CELL_CHANGED_TIMEOUT = 1000;
 export class CellManager implements IDisposable {
   constructor(options: CellManager.IOptions) {
     this._debuggerService = options.debuggerService;
+    this._id = options.debuggerService.session.client.name;
```
jtpio: Is this `_id` supposed to the unique? `client.name` can be the same for 2 notebooks (with different sessions) in 2 different folders (named `Untitled.ipynb` for example).

src/handlers/cell.ts
jtpio: OK that's the previous one that was moved around. Let's keep it for now.

src/handlers/cell.ts
jtpio: Maybe we could create an id on the debug session.

jtpio: There is another `id` comparison happening in the notebook handler:

https://github.com/jupyterlab/debugger/blob/2c4a4ca854581aa9138686c7c6b9fc7defb71bd4/src/handlers/notebook.ts#L64

Maybe these two could be combined into one. Will look at it separately.

jtpio:
https://github.com/jupyterlab/debugger/issues/204 should be fixed.

But it looks like the click handlers are still set:

debugger-disposed

jtpio:
I left a `TODO` so we can look at whether we need to track an id for the `CellManager` and if so which one.

PR #193, opened 2019-11-15
Add the sidebar to the layout restorer

### LINKED ISSUE #96: Restore open sidebar
At the moment if we open the sidebar and then reload the page, the sidebar will be minimized:

restore-sidebar-state

It would be great if the sidebar could keep its "open" state on reload.

jtpio:
This should fix #96.

Add the sidebar to the layout restorer so it can be restored in the right area on page reload, and serializes it in the workspace file:

```json
"right": {
	"collapsed": false,
	"current": "debugger-sidebar",
	"widgets": [
	  "debugger-sidebar"
	]
}
```

restore-sidebar-state2

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -277,6 +277,10 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
         sidebar.title.label = 'Environment';
         shell.add(sidebar, 'right', { activate: false });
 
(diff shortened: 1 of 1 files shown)


KsavinN:
checked locally work well. It fixes #96

Issue #166, opened 2019-11-08
Add a test for DebugService.restart


jtpio:
We should have a test for the `restart` method in the `DebugService`.

Useful for changes like: https://github.com/jupyterlab/debugger/pull/165

jtpio:
Fixed by #198.

PR #198, opened 2019-11-18
Added tests for restoreState and updateBreakpoints

JohanMabille:


tests/src/service.spec.ts
```diff
@@ -91,4 +93,58 @@ describe('DebugService', () => {
...
+            path: source_id
+          }
+        };
+      });
+      await service.updateBreakpoints(code, breakpoints);
+    });
+
+    describe('#updateBreakpoints', () => {
+      it('should update the breakpoints', () => {
+        const bp_list = model.breakpointsModel.getBreakpoints(source_id);
```
jtpio: camelCase for the naming of variables? :)

JohanMabille: I did too much C++ during the weekend

JohanMabille: I did too much C++ during the weekend

jtpio:
@JohanMabille could this change also include https://github.com/jupyterlab/debugger/issues/166? or should this be done separately?

JohanMabille:
Ah good catch. I'll add it.

src/tokens.ts
```diff
@@ -85,7 +85,7 @@ export interface IDebugger extends IDisposable {
 
   /**
    * Restart the debugger.
-   * Precondition: isStarted()
+   * Precondition: !isStarted()
```
jtpio: Is that true?

JohanMabille: No, that's a wrong update.

src/service.ts
```diff
@@ -194,12 +194,10 @@ export class DebugService implements IDebugger {
     // No need to dump the cells again, we can simply
     // resend the breakpoints to the kernel and update
     // the model.
-    breakpoints.forEach(async (bp, path, _) => {
-      const sourceBreakpoints = Private.toSourceBreakpoints(bp);
-      await this.setBreakpoints(sourceBreakpoints, path);
-    });
-
-    this.model.breakpointsModel.restoreBreakpoints(breakpoints);
+    for (const bp of breakpoints) {
```
jtpio: How about something like this?

```suggestion
    for (const [source, bps] of breakpoints) {
```

JohanMabille: definitely better!

jtpio:
This fixes #166.

PR #199, opened 2019-11-18
Initial support for handling messages for other cells

### LINKED ISSUE #121: Executing via the Run button
Executing a code cell via Ctrl-Enter seems to be working fine (cf #120).

But clicking on the run button advances to the next cell before the execution has finished:

run-button-bug

jtpio:
This will fix #121.

There will still be more to refactor afterwards. Ideally we should be able to jump to any cells that are tracked by the debugger plugins. The logic could be shared between the notebook and console handlers.

Opening the content of the file / cell when it cannot be found in the current widgets can be done in #169.

### TODO

- [x] Find the cell where execution has stopped using its hash
- [x] Properly cleanup the cell on debug actions (continue)
- [x] Bring focus back to the cell where execution has stopped

run-button-jump-to-cell

### DIFF
--- a/src/handlers/cell.ts
+++ b/src/handlers/cell.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { CodeCell, ICellModel } from '@jupyterlab/cells';
+import { Cell, CodeCell, ICellModel } from '@jupyterlab/cells';
 
 import { CodeMirrorEditor } from '@jupyterlab/codemirror';
 
@@ -14,6 +14,8 @@ import { Signal } from '@phosphor/signaling';
 import { Editor } from 'codemirror';
 
 import { Breakpoints, SessionTypes } from '../breakpoints';
+
+import { Callstack } from '../callstack';
 
 import { Debugger } from '../debugger';
 
@@ -43,11 +45,10 @@ export class CellManager implements IDisposable {
 
     this._debuggerModel.callstackModel.currentFrameChanged.connect(
       (_, frame) => {
-        this.cleanupHighlight();
+        CellManager.cleanupHighlight(this.activeCell);
         if (!frame) {
           return;
         }
-        this.showCurrentLine(frame.line);
       }
     );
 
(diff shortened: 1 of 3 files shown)


jtpio:
Instead of cells, we could start speaking in terms of `editors`.

The debugger keeps track of the editors (cells, console prompts, files) and would iterate over them to determine the one that matches the source coming back from the DAP.

If it doesn't find the editor, we'll use the tab panel in the main area widget to show a read-only copy of the source (#169).

jtpio:
`editors` used to be part of the interface at some point (but can't find anymore at the moment).

jtpio:
Update:

run-button-jump-to-cell

src/handlers/cell.ts
```diff
@@ -291,6 +272,22 @@ export namespace CellManager {
     activeCell?: CodeCell;
     type: SessionTypes;
   }
+
+  export function showCurrentLine(cell: Cell, frame: Callstack.IFrame) {
```
jtpio: These should have docstrings.

jtpio: Fixed.

jtpio:
This means we can also jump between cells like this:

jump-to-cell

jtpio:
> `editors` used to be part of the interface at some point (but can't find anymore at the moment).

This can be tracked in #200 and handled separately.

src/handlers/notebook.ts
```diff
@@ -63,6 +70,34 @@ export class NotebookHandler implements IDisposable {
...
+    cells.forEach(cell => CellManager.cleanupHighlight(cell));
+
+    if (!frame) {
+      return;
+    }
+
+    cells.forEach((cell, i) => {
+      // check the event is for the correct cell
+      const code = cell.model.value.text;
+      const cellId = this.debuggerService.getCellId(code);
```
JohanMabille: Attached properties would avoid recomputing the cell id (although this can be done later).
Also I wonder if it would be worth maintaining a mapping CellId <-> Cells so we don't have to for through all the cells when a frameChanged event is fired.

jtpio: > Attached properties would avoid recomputing the cell id (although this can be done later).

But the id / hash would still need to be computed when the cell changes.

> Also I wonder if it would be worth maintaining a mapping CellId <-> Cells so we don't have to for through all the cells when a frameChanged event is fired.

That sounds like a good idea. Do you think this is something we could do as part of #200?

JohanMabille: Yes.

JohanMabille:
LGTM, let's merge!

PR #201, opened 2019-11-18
Add support for the debugger flag in kernel info

### LINKED ISSUE #167: Check for kernel debugging capabilities
We should check whether a kernel has debugging capabilities or not before starting the debugger.

Related `TODO`: https://github.com/jupyterlab/debugger/blob/f64127073f51361cc48d76ce95d6ff93ba0544ee/src/service.ts#L28-L29

There is a PR to add the `debugger` flag for `kernel_info_reply` in `jupyter_client`: https://github.com/jupyter/jupyter_client/pull/486

In the meantime we can create an interface that extends [`IInfoReply`](https://github.com/jupyterlab/jupyterlab/blob/2eca1c67f7f03c9d04c1ecc205f7a32448e37300/packages/services/src/kernel/messages.ts#L755-L762) and adds the `debugger` flag.

jtpio:
Fixes #167.

### TODO

- [x] Add `debugger` key to the interface
- [x] ~Check debugging is enabled before starting the service~

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -80,7 +80,9 @@ async function setDebugSession(
   } else {
     debug.session.client = client;
   }
-  await debug.restoreState(true);
+  if (debug.isDebuggingEnabled) {
+    await debug.restoreState(true);
+  }
   app.commands.notifyCommandChanged();
 }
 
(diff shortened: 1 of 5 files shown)


JohanMabille:
LGTM!

jtpio:
Whether we should show the debugger or not can be handled separately.

jtpio:
> LGTM!

Taking this LGTM as approval :)

More to come in follow-up PRs.

PR #203, opened 2019-11-19
Use the value.changed signal for the cell monitor

### LINKED ISSUE #202: Extra dumpCell request sent
After a breakpoint is hit and the execution is resumed with `Next` until it reaches the end, an extra `dumpCell` request is sent to the backend. From the callstack, the root cause seems to be the activity monitor:

```typescript
handlers/cells.ts: 156
      this._cellMonitor.activityStopped.connect(() => {
        this.sendEditorBreakpoints();
      }, this);
```

jtpio:
Fixes #202.

The `value.changed` signal seems more appropriate.

value-changed-signal

### DIFF
--- a/src/handlers/cell.ts
+++ b/src/handlers/cell.ts
@@ -1,11 +1,13 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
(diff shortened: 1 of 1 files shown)


jtpio:
cc @JohanMabille, this should fix #202.

JohanMabille:
Thanks!

PR #206, opened 2019-11-19
Upgraded to xeus-python 0.6.7

JohanMabille:


jtpio:
There seems to be an increase of failed tests recently (compared to before).

For example: https://dev.azure.com/jupyterlab/jupyterlab/_build/results?buildId=4059

Has anything in xeus-python changed that could be related to this?

That might also just be a coincidence (or flaky tests).

JohanMabille:
This looks like flaky tests, tests were passing before I merge the PR (on both the CI and my machine).

jtpio:
OK cool, will keep an eye on it :+1:

Issue #200, opened 2019-11-18
Replace CellManager by EditorManager?


jtpio:
At the moment the logic revolves around the notion of a `Cell`.

However (and I think it was the case during the initial iterations on the interfaces), we should probably start thinking in terms of editors instead.

The Debugger would keep track of a list of `CodeEditor.IEditor`. An editor can be a code cell, but also a console prompt or a read-only editor in the main area widget.

This could simplify the logic when the debugger needs to find which editor corresponds to a debug event (for example `stopped`).

jtpio:
This is relevant for https://github.com/jupyterlab/debugger/issues/169

jtpio:
Some parts are covered in #207, but not everything.

PR #207, opened 2019-11-20
Open files in the main area widget

### LINKED ISSUE #169: Open files in the tab panel editors widget

jtpio:
Fixes #169.

### Steps

- [x] Basic support for opening files in the main area widget
- [x] Rename `cell.ts` to `editor.ts` and `CellManager` to `EditorHandler`
- [x] Bring focus to the right file when stepping or opening a new file

read-only-editors


jtpio:
Wondering whether we should still keep this idea of an "active cell":

https://github.com/jupyterlab/debugger/blob/2c4a4ca854581aa9138686c7c6b9fc7defb71bd4/src/handlers/cell.ts#L35

We could still keep the handler on `activeCellChanged` to create a new `CellManager` (or rather `EditorHandler`?), to wire everything when a new cell is being edited for the current debug session.

src/tokens.ts
```diff
@@ -130,6 +130,14 @@ export interface IDebugger extends IDisposable {
...
    */
   clearBreakpoints(): Promise<void>;
+
+  /**
+   * Retrieve the content of a source file.
+   * @param source The source object containing the path to the file.
+   */
+  getSource(
+    source: DebugProtocol.Source
+  ): Promise<DebugProtocol.SourceResponse>;
```
jtpio: This could instead return a different interface than `DebugProtocol.SourceResponse`, that would only contain `content` and `mimeType`.

jtpio:
Since this PR can become quite big, the remaining tasks have been split into several issues:

- https://github.com/jupyterlab/debugger/issues/200
- https://github.com/jupyterlab/debugger/issues/211
- https://github.com/jupyterlab/debugger/issues/212

And marking this one as ready for review.

jtpio:
Demo when stepping in a local used-defined module:

local-module

src/tokens.ts
JohanMabille: I agree with this.

JohanMabille: This can be done in another PR, though

Issue #214, opened 2019-11-20
Should the debugger lifecycle be managed by a toggleable button?


afshin:
- We should only make a debugger button/toggle available in the toolbar of consoles and notebooks that have kernels that handle debugging.
- If we make the "button" actually be a toggle/switch/checkbox which indicates whether the kernel in question has currently started or stopped debugging, then it could allow users to switch between notebooks and have the debugger UI *only* show the last *active/started* debuggable kernel's state.
  - This means that if you open up notebook **foo.ipynb** and start its debugger, the debugger UI appears and tracks notebook **foo.ipynb**'s kernel state.
  - If you then open up notebook **bar.ipynb** and *do not* toggle its debugger, then you can still see the state of notebook **foo.ipynb** even if the active notebook that has the focus is notebook **bar.ipynb**.

cf. https://github.com/jupyterlab/debugger/issues/180

jtpio:
Fixed by #239.

PR #216, opened 2019-11-21
no create handle for no supported kernels

### LINKED ISSUE #209: Breakpoint marker hover for kernels without debugging support
At the moment we show the breakpoint markers also for kernels that do not support debugging (for example `ipykernel`) if the line numbers are toggled on:

breakpoint-gutter-hover-ipykernel

If we don't want the debugger to interfere with anything until it's open, we should look into a way to not show these markers (maybe using the changes from https://github.com/jupyterlab/debugger/pull/201).

KsavinN:
It's resolve #209 . 

But seems like during creating new widget client not have already attached kernel  (it's exist some time gap):
issue_on_start

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -95,7 +95,7 @@ class DebuggerHandler<H extends ConsoleHandler | NotebookHandler> {
     T extends IConsoleTracker | INotebookTracker,
     W extends ConsolePanel | NotebookPanel
   >(debug: IDebugger, tracker: T, widget: W): void {
-    if (!debug.model || this.handlers[widget.id]) {
+    if (!debug.model || this.handlers[widget.id] || !debug.isDebuggingEnabled) {
       return;
     }
     const handler = new this.builder({


jtpio:
> But seems like during creating new widget client not have already attached kernel (it's exist some time gap)

Yes we should probably check the kernel session is ready first.

If I recall correctly the service has an `ensureSessionReady` method.

KsavinN:
After @jtpio fixes #219  it seems work good
isWorking

JohanMabille:
@KsavinN I think you should rebase your branch on the current master, that would help reading the diffs and will avoid you too many conflicts if other PRs are merged.

src/index.ts
```diff
@@ -95,26 +95,26 @@ class DebuggerHandler<H extends ConsoleHandler | NotebookHandler> {
...
-    const handler = new this.builder({
-      debuggerService: debug,
-      widget
-    });
-    this.handlers[widget.id] = handler;
-    widget.disposed.connect(() => {
-      handler.dispose();
-      delete this.handlers[widget.id];
-    });
+    if (debug.model && !this.handlers[widget.id] && debug.isDebuggingEnabled) {
```
jtpio: This reverts the change from https://github.com/jupyterlab/debugger/pull/221/commits/226e55a9127533cd0f3a4f1a6eb4e09c0574d2a7.

KsavinN: good catch. I missed that during make rebase

jtpio:
We'll probably want to iterate on that a bit more when we start looking at https://github.com/jupyterlab/debugger/issues/214.

PR #217, opened 2019-11-21
Restores the stopped threads after a restoreState request

### LINKED ISSUE #205: Restore the full state
We should now be able to restore the full state of the current execution using the `stoppedThreads` field from the `debugInfo` reply:

https://github.com/jupyterlab/debugger/blob/b1b41742d13aab5c6243d7fca0c66f91ac288ba9/src/tokens.ts#L231

The debugger will then need to request the state of variables for each of the threads.

JohanMabille:
Fixes #205

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -158,6 +158,20 @@ export namespace Debugger {
       this._modeChanged.emit(mode);
     }
 
+    /**
+     * The set of threads in stopped state.
+     */
+    get stoppedThreads(): Set<number> {
+      return this._stoppedThreads;
+    }
+
+    /**
+     * Assigns the parameters to the set of threads in stopped state.
+     */
+    set stoppedThreads(threads: Set<number>) {
+      this._stoppedThreads = threads;
+    }
+
     get modeChanged(): ISignal<this, IDebugger.Mode> {
       return this._modeChanged;
     }
@@ -181,6 +195,7 @@ export namespace Debugger {
     private _codeValue: IObservableString;
     private _isDisposed = false;
     private _mode: IDebugger.Mode;
+    private _stoppedThreads = new Set<number>();
     private _modeChanged = new Signal<this, IDebugger.Mode>(this);
     private _disposed = new Signal<this, void>(this);
   }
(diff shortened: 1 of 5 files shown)


src/service.ts
```diff
@@ -165,10 +165,10 @@ export class DebugService implements IDebugger {
   }
 
   /**
-   * Whether the current thread is stopped.
+   * Whether there exist a thread in stopped stated.
```
jtpio: Typo?

```suggestion
   * Whether there exist a thread in a stopped state.
```

src/service.ts
```diff
@@ -165,10 +165,10 @@ export class DebugService implements IDebugger {
...
    */
-  isThreadStopped(): boolean {
-    return this._stoppedThreads.has(this.currentThread());
+  hasThreadStopped(): boolean {
```
jtpio: Or maybe `hasStoppedThread()`?

src/debugger.ts
```diff
@@ -156,6 +156,14 @@ export namespace Debugger {
       this._modeChanged.emit(mode);
     }
 
+    get stoppedThreads(): Set<number> {
```
jtpio: Let's also add a docstring?

src/service.ts
JohanMabille: yep that would be consistent with the name of the variable

jtpio:
This is what it looks like:

restore-stopped-threads

jtpio:
But reloading the page:

restore-stopped-threads-reload

JohanMabille:
The issue relative to reloading the page is deeper and requires that we take some decision regarding the different options we have to handle it.
I will open an issue to give more detail about that.

src/service.ts.orig
```diff
@@ -0,0 +1,543 @@
+// Copyright (c) Jupyter Development Team.
```
jtpio: Should we add these `.orig` to the `.gitignore`?

JohanMabille: Yes we can do that. I'm going to remove the .orig added by mistake, I'll add the gitignore change in a dedicated commit.

PR #221, opened 2019-11-22
Add support for breakpoints and line highlighting in read-only editors

### LINKED ISSUE #200: Replace CellManager by EditorManager?

### LINKED ISSUE #211: Add support for breakpoints in the read-only editors
Just like for cells, we should be able to add and remove breakpoints for the read-only editors.

image

jtpio:
Fixes #200 
Fixes #211 

The goal is to make the `EditorHandler` aware of `CodeEditor.IEditor`, so we can reuse the same logic and hooks (gutter click, line highlighting and so on), when opening new editors in the main area widget.

### Changes

- Remove `codeValue` from `Debugger.Model`
- `EditorHandler` handles a `CodeEditor.IEditor` instead of a `Cell`
- Remove notebook and console trackers from the handlers (a handler is for one widget)
- Re-use the `EditorHandler` for the main area read-only editors: add support for breakpoints and line highlighting

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -5,8 +5,6 @@ import { IEditorServices } from '@jupyterlab/codeeditor';
 
 import { IDataConnector } from '@jupyterlab/coreutils';
 
-import { IObservableString } from '@jupyterlab/observables';
-
 import { ReadonlyJSONValue } from '@phosphor/coreutils';
 
 import { Message } from '@phosphor/messaging';
@@ -174,14 +172,6 @@ export namespace Debugger {
       return this._modeChanged;
     }
 
-    get codeValue() {
-      return this._codeValue;
-    }
-
-    set codeValue(observableString: IObservableString) {
-      this._codeValue = observableString;
-    }
-
     private async _populate(): Promise<void> {
       const { connector } = this;
 
@@ -190,7 +180,6 @@ export namespace Debugger {
       }
     }
 
-    private _codeValue: IObservableString;
     private _isDisposed = false;
     private _mode: IDebugger.Mode;
     private _stoppedThreads = new Set<number>();
--- a/src/editors/index.ts
+++ b/src/editors/index.ts
@@ -30,6 +30,10 @@ export class DebuggerEditors extends TabPanel {
 
     this.tabsMovable = true;
     this.tabBar.insertBehavior = 'select-tab';
+    this.tabBar.tabCloseRequested.connect((_, tab) => {
+      const widget = tab.title.owner;
+      widget.dispose();
+    });
 
     this.model = new DebuggerEditors.IModel();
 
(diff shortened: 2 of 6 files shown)


jtpio:
This is still a work in progress at the current moment.

jtpio:
Reusing the `EditorHandler` for the main area read-only editors:

main-area-editors-highlight

jtpio:
Now syncing the breakpoints correctly:

main-area-editors-bps

The highlight is still removed when switching cells though.

jtpio:
This is what it looks like now without the flickering of the breakpoints:

main-area-editors-bps2

jtpio:
This should be ready now.

Since there is quite a bit of refactoring and cleanup, it would be nice if someone else could also test the changes locally, or on Binder:

[Binder](https://mybinder.org/v2/gh/jtpio/debugger/editor-breakpoints?urlpath=/lab/tree/examples/index.ipynb)

src/debugger.ts
```diff
@@ -174,14 +172,6 @@ export namespace Debugger {
...
     }
 
-    get codeValue() {
-      return this._codeValue;
-    }
-
-    set codeValue(observableString: IObservableString) {
-      this._codeValue = observableString;
-    }
-
```
JohanMabille: Nice! It's been a while that I wanted to remove this from the model, I'm glad you did it!

JohanMabille:
After some tests, I've encountered some issues (sorry for not posting a screencast, I have some issues doing it on thi smachine). After hitting the breakpoint on the `add` as in your first screencast, if I select the cell where `foo` is defined, I observe randomly the following behaviors:

- either the front end forces the focus back to the cell were the code has stopped (that is, the one containing `add`) and the highlight is correctly restored in the notebook.
- or nothing happens, and when I select the cell containing `add`, the highlight is not restored in the cell.

In both cases, the higlight is correct in the file panel.

jtpio:
>  * either the front end forces the focus back to the cell were the code has stopped (that is, the one containing `add`) and the highlight is correctly restored in the notebook.

Yes the notebook does change the selected cell when a frame is selected to focus where the stopped event happened.

The reason it seems to be random might be because when the notebook gets back the focus, everything is triggered again with the restore including setting the index of the active cell.

This "jump" also happens when switching from the main area widget to the notebook (since the current widget is changing):

notebook-cell-jump-focus

> * or nothing happens, and when I select the cell containing `add`, the highlight is not restored in the cell.

This one is because the `EditorHandler` is disposed when switching cells, and the highlight is cleared in the dispose. We can fix this by not clearing the highlight in the dispose, since it's already cleared in the notebook handler anyway.

JohanMabille:
Thanks!

PR #222, opened 2019-11-22
Imodel

### LINKED ISSUE #218: Add a interface for Debugger.Model
Currently IDebugger exposes the Debugger.Model class in its API while it should expose an interface only.

JohanMabille:
Fixes #218 
The interface `IModel` does not expose anything for now, to limit the constraints on the API if this get integrated in JupyterLab 2.0.

### DIFF
--- a/src/tokens.ts
+++ b/src/tokens.ts
@@ -16,7 +16,6 @@ import { DebugProtocol } from 'vscode-debugprotocol';
 // TODO: remove that import when an interface has
 // been created for Model class
 import { Breakpoints } from './breakpoints';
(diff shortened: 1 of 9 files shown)


src/tokens.ts
```diff
@@ -205,6 +200,18 @@ export namespace IDebugger {
     ): Promise<IDebugger.ISession.Response[K]>;
   }
 
+  /**
+   * Single breakpoint in a cell.
```
jtpio: Or "in an editor"?

JohanMabille: good catch

Issue #176, opened 2019-11-13
Alternative to the main area widget tab panel


jtpio:
At the moment the main area debugger widget contains both the sidebar and the tab panel for read-only editors:

![image](https://user-images.githubusercontent.com/591645/68753126-c0a69580-0604-11ea-8d30-16835c5ef141.png)

This is according to the current [design](https://github.com/jupyterlab/debugger/blob/ffdd985e69247d9bc39249d98ff57c44b6d84cb8/DESIGN.md) document. However we could also consider the following:

- keep only one mode: `condensed` (= sidebar). This will simplify the restore flow and remove the need for switching between modes
- when we need to display read-only files, we would create a new read-only editor and add it to the `DockPanel` in the main area (next to the notebooks and other activities)
- we could decide whether to close the read-only editors when all threads have continued, or leave them in the main area and let the user close them manually

This would change the behavior to be more similar to existing IDEs where the debugger sidebar is usually attached by default (left or right), and files are open in the main area when stepping into code.

jtpio:
Also related is what should happen when the activity (notebook, console) that currently has focus doesn't support debugging. For example a notebook with `ipykernel` as the kernel.

If there is a main area widget currently open, and the user switches to a notebook that doesn't support debugging, then the main area widget would need to be closed (or entirely disabled). When switching back it will have to be open (or re-enabled) in the main area. Could this create confusion?

With 1 mode only, only the sidebar would have to be removed when debugging is not possible.

Related to https://github.com/jupyterlab/debugger/issues/181

jtpio:
Pasting the mock-up from #229 here for reference:

![image](https://user-images.githubusercontent.com/591645/69960507-f2c35d00-1509-11ea-801b-1f578dfecfa1.png)

Instead of `CODE` the extra `Panel` could say `SOURCE` instead.

PR #226, opened 2019-11-25
Add "Change Mode" to the sidebar context menu

### LINKED ISSUE #178: Add a button (or menu item) to switch between modes
This has been mentioned several times but there doesn't seem to be any issue tracking it.

If we want to be able to switch modes, there should be a button or some UI element to do so, that will map to the `changeMode` command.

However this might not be needed if we stick to 1 mode, see https://github.com/jupyterlab/debugger/issues/176.

jtpio:
Fixes #178.

Simple fix to make it easier to change modes without having to use the palette.

We might want to add a real button later on, or something that make it easier to discover.

image

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -263,6 +263,11 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
     });
 
     app.contextMenu.addItem({
+      command: CommandIDs.changeMode,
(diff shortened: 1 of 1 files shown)


jtpio:
This is also ready for review.

Although there is an open issue that proposes keeping only one mode: #176.

PR #228, opened 2019-11-25
Remove deactivate breakpoint

### LINKED ISSUE #213: Remove the "Deactivate Breakpoint" button?
This is something to consider for the first release.

If we decide to keep #128 for later, we should then remove the UI elements that could let users think it is possible to deactivate breakpoints:

- the "Deactivate Breakpoint" button
- the checkboxes (can be replaced by the `jp-CircleIcon` for example)

image

KsavinN:
resolved #213 

Backend, for now, don't support deactivate breakpoints. We just remove (commended) input checkbox and remove toolbar from breakpoints panel 'deactive all breakpoints'.

chrome-capture

### DIFF
--- a/src/breakpoints/index.ts
+++ b/src/breakpoints/index.ts
@@ -25,25 +25,6 @@ export class Breakpoints extends Panel {
     this.addWidget(header);
     this.addWidget(this.body);
 
-    // header.toolbar.addItem(
-    //   'deactivate',
-    //   new ToolbarButton({
-    //     iconClassName: 'jp-DebuggerDeactivateIcon',
-    //     tooltip: `${this.isAllActive ? 'Deactivate' : 'Activate'} Breakpoints`,
-    //     onClick: () => {
-    //       this.isAllActive = !this.isAllActive;
-
-    //       // TODO: this requires a set breakpoint(bp: Breakpoints.IBreakpoint[]) method in the model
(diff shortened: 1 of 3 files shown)


src/breakpoints/index.ts
```diff
@@ -25,24 +25,24 @@ export class Breakpoints extends Panel {
...
-          /*Array.from(this.model.breakpoints.values()).map((breakpoints: Breakpoints.IBreakpoint[]) => {
-            breakpoints.map((breakpoint: Breakpoints.IBreakpoint) => {
-              breakpoint.active = this.isAllActive;
-              this.model.breakpoint = breakpoint;
-            });
-          });*/
-        }
-      })
-    );
+    // header.toolbar.addItem(
```
jtpio: Should we just remove the code instead of commenting it?

jtpio: This is tracked in the git history, so we could still find the previous code.

We can also keep a reference to the commit to make it easier to find it back later on.

jtpio:
Should we also remove the `breakpointChanged` signal?

https://github.com/jupyterlab/debugger/blob/57951d6cfb065f16c631a281fca1dfb77f0d0b09/src/breakpoints/index.ts#L95-L98

I think it was added to handle activating and deactivating breakpoints. We could remove it for now since it is not used anyway.

Issue #224, opened 2019-11-25
Clickable breakpoints


jtpio:
It would be useful to make the breakpoints clickable (similar to the stack frames).

Clicking on a breakpoint would bring the focus back on the editor where it is defined (for example it would jump back to the cell where it is defined).

For cells that were executed before but that do not show up in the UI, we could open the main area widget and a read-only editor.

jtpio:
This could actually be the same logic as when an editor is open in the main area widget (https://github.com/jupyterlab/debugger/issues/212, not implemented yet).

There could be a function that takes a source path (and an optional line number) as input, and goes through the list of editors tracked for the current debug session. It would stop once there is a match (by comparing the hash of the editor content). If there is no match it opens the file in the main area widget.

PR #231, opened 2019-11-26
Add support for debugging files

### LINKED ISSUE #225: Debug files
Now that the `EditorHandler` tracks editors, we should be able to reuse it to debug files too.

We'll also need a button to execute the code. That button could be added to the toolbar of the file editor.

jtpio:
Fixes #225.

For now this still requires having a console open, and pressing `Shift-Enter` to execute the selected code.

Later on we will want to add a play or debug button to debug the file. 

### Changes

- [x] The Debug session can take a `IClientSession` or `Session.ISession`
- [x] Add `FileHandler` to handler `FileEditor` widgets (reusing the `EditorHandler`)
- [x] Show line highlighting in the `FileEditor` editor
- [x] Rename `getCellId` to `getCodeId`
- [x] Handle restore

debug-files2

### DIFF
--- a/src/handlers/file.ts
+++ b/src/handlers/file.ts
@@ -56,8 +56,8 @@ export class FileHandler implements IDisposable {
     }
 
     const code = editor.model.value.text;
-    const cellId = this.debuggerService.getCodeId(code);
-    if (frame.source.path !== cellId) {
+    const codeId = this.debuggerService.getCodeId(code);
+    if (frame.source.path !== codeId) {
       return;
     }
     // request drawing the line after the editor has been cleared above
--- a/src/handlers/file.ts
+++ b/src/handlers/file.ts
@@ -56,7 +56,7 @@ export class FileHandler implements IDisposable {
     }
 
     const code = editor.model.value.text;
-    const cellId = this.debuggerService.getCellId(code);
+    const cellId = this.debuggerService.getCodeId(code);
     if (frame.source.path !== cellId) {
       return;
     }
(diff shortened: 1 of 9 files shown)


jtpio:
Now handling the session restore:

debug-files3

src/handlers/file.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const code = editor.model.value.text;
+    const cellId = this.debuggerService.getCodeId(code);
+    if (frame.source.path !== cellId) {
+      return;
+    }
+    // request drawing the line after the editor has been cleared above
+    requestAnimationFrame(() => {
+      EditorHandler.showCurrentLine(editor, frame);
+    });
```
jtpio: This logic is similar to the one in the notebook handler.

We could instead have one handler that tracks the editors for the current debug session, and finds the one that matches the hash.

(to be done in a different change, similar to #224)

src/index.ts
```diff
@@ -160,36 +164,36 @@ const files: JupyterFrontEndPlugin<void> = {
...
-        const session = sessions.connectTo(model);
-        debug.session.client = session;
-      });
-    });
-
-    app.commands.addCommand(CommandIDs.debugFile, {
-      execute: async _ => {
-        if (!tracker || !tracker.currentWidget) {
-          return;
+      void sessions.findByPath(widget.context.path).then(async model => {
```
jtpio: I think this doesn't handle the case when there is no session for the given path (not backed by a kernel).

Will submit a fix to handle that.

jtpio: Done.

PR #232, opened 2019-11-26
no throws error after close debugger

### LINKED ISSUE #227: Closing the debugger when there is no active notebook throws an error
Closing the debugger when there is no active notebook throws an error.

This seems to be because we call the `debug.stop()` method when the model is disposed:

https://github.com/jupyterlab/debugger/blob/57951d6cfb065f16c631a281fca1dfb77f0d0b09/src/index.ts#L112

Do we need to stop the debugger?

To reproduce:

1. Open the debugger
2. Open a notebook
3. Close the notebook
4. Close the debugger

debugger-close-no-notebook

KsavinN:
Resolve #227

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -109,8 +109,12 @@ class DebuggerHandler<H extends ConsoleHandler | NotebookHandler> {
     });
 
     debug.model.disposed.connect(async () => {
+      const handlerIds = Object.keys(this.handlers);
+      if (handlerIds.length === 0) {
+        return;
+      }
       await debug.stop();
-      Object.keys(this.handlers).forEach(id => {
+      handlerIds.forEach(id => {
         this.handlers[id].dispose();
       });
       this.handlers = {};


src/index.ts
```diff
@@ -109,11 +109,14 @@ class DebuggerHandler<H extends ConsoleHandler | NotebookHandler> {
     });
 
     debug.model.disposed.connect(async () => {
-      await debug.stop();
-      Object.keys(this.handlers).forEach(id => {
-        this.handlers[id].dispose();
-      });
-      this.handlers = {};
+      const idsOfHandler = Object.keys(this.handlers);
```
jtpio: How about `handlerIds` instead of `idsOfHandler`?

src/index.ts
```diff
@@ -109,11 +109,14 @@ class DebuggerHandler<H extends ConsoleHandler | NotebookHandler> {
...
+      if (idsOfHandler.length > 0) {
+        await debug.stop();
+        idsOfHandler.forEach(id => {
+          this.handlers[id].dispose();
+        });
+        this.handlers = {};
+      }
```
jtpio: How about using an early return if there is no handler?

```typescript
if (handlerIds.length === 0) {
  return;
}
await debug.stop();
handlerIds.forEach(id => {
  this.handlers[id].dispose();
});
this.handlers = {};

```

src/index.ts
KsavinN: Resolved

src/index.ts
KsavinN: Resolved

PR #233, opened 2019-11-26
set data atrribute

### LINKED ISSUE #137: Add a data attribute to notebooks/consoles that support debugging.
We should use a mechanism like this to make sure that classes we add (for example to support gutter hovers, etc.) are only visible in notebooks and consoles where they are relevant. We could then target them via CSS selectors, e.g.:

```css
[data-jp-debugger='true'] .CodeMirror-gutter-wrapper::after {
  content: '●';
  color: var(--jp-error-color1);
  /* ... */
}
```

KsavinN:
resolved #137

### DIFF
--- a/src/editors/index.ts
+++ b/src/editors/index.ts
@@ -29,6 +29,7 @@ export class DebuggerEditors extends TabPanel {
     super();
     this.tabsMovable = true;
     this.tabBar.insertBehavior = 'select-tab';
+    this.node.setAttribute('data-jp-debugger', 'true');
     this.tabBar.tabCloseRequested.connect((_, tab) => {
       const widget = tab.title.owner;
       widget.dispose();
@@ -40,6 +41,14 @@ export class DebuggerEditors extends TabPanel {
     this.debuggerService = options.service;
     this.editorFactory = options.editorServices.factoryService.newInlineEditor;
     this.mimeTypeService = options.editorServices.mimeTypeService;
+
+    this.debuggerService.session?.clientChanged.connect((_, updated) => {
+      if (this.tabBar.currentTitle?.owner.id === updated.path) {
+        this.node.setAttribute('data-jp-debugger', 'true');
+      } else {
+        this.node.setAttribute('data-jp-debugger', 'false');
+      }
+    });
 
     this.tabBar.currentChanged.connect((_, updated) => {
       if (
--- a/src/editors/index.ts
+++ b/src/editors/index.ts
@@ -27,7 +27,6 @@ import { IDebugger } from '../tokens';
 export class DebuggerEditors extends TabPanel {
   constructor(options: DebuggerEditors.IOptions) {
     super();
-
     this.tabsMovable = true;
     this.tabBar.insertBehavior = 'select-tab';
     this.tabBar.tabCloseRequested.connect((_, tab) => {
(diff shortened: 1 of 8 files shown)


src/editors/index.ts
```diff
@@ -42,6 +42,25 @@ export class DebuggerEditors extends TabPanel {
     this.editorFactory = options.editorServices.factoryService.newInlineEditor;
     this.mimeTypeService = options.editorServices.mimeTypeService;
 
+    this.debuggerService.session?.clientChanged.connect((_, updated) => {
+      if (this.tabBar?.currentTitle?.owner?.id === updated.path) {
```
jtpio: We probably don't need to use the optional chaining with `?` everywhere.

For example here `this.tabBar` should not be `null`.

src/session.ts
```diff
@@ -64,9 +64,14 @@ export class DebugSession implements IDebugger.ISession {
 
     if (client) {
       this._client.iopubMessage.connect(this._handleEvent, this);
+      this._clientChanged.emit(client as IClientSession);
```
jtpio: This PR would need to be rebased to take into account the changes from #231, which adds support for `Session.ISession`.

src/editors/index.ts
KsavinN: tabBar exist before. I will just check if currentTitle exist.

src/session.ts
KsavinN: resolved.

src/editors/index.ts
KsavinN: for session i put optional chaining cause test are not passed

src/editors/index.ts
```diff
@@ -134,6 +152,7 @@ export class DebuggerEditors extends TabPanel {
     editor.title.label = path;
     editor.title.caption = path;
     editor.title.closable = true;
+    editor.id = this.debuggerService.session.client.path;
```
jtpio: Wondering whether we need an `id`, or we could use `editor.title.label`?

src/handlers/console.ts
```diff
@@ -20,6 +20,7 @@ export class ConsoleHandler implements IDisposable {
     this.debuggerModel = options.debuggerService.model as Debugger.Model;
     this.debuggerService = options.debuggerService;
     this.consolePanel = options.widget;
+    this.consolePanel.node.setAttribute('data-jp-debugger', 'true');
```
jtpio: Should we set the attributes in both the notebook / console handlers and in the editor handler? Or would one place be enough?

src/editors/index.ts
KsavinN: Due to we want always to have the possibility to set breakpoints even for not focused editors I removed a connection to clienteChanged and tabBar changed.

jtpio: Then do we need to keep the `clientChanged` signal?

KsavinN: I removed that

jtpio: :+1:

src/editors/index.ts
jtpio: After the rebase it looks like it can actually be removed.

KsavinN: Removed.

jtpio: :+1:

Issue #212, opened 2019-11-20
Automatically switch to expanded mode when stepping into code


jtpio:
When the current mode is condensed (sidebar) and we step into:

- foreign code
- a custom module that lives in a different file
- a function that was defined in a cell that has been deleted
- or anything that is not visually open in the workspace

the debugger should switch to expanded mode and open the right file in the editors area.

jtpio:
This might not be needed, and depends on #176.

jtpio:
Closing in favor of https://github.com/jupyterlab/debugger/issues/176.

Issue #230, opened 2019-11-26
After close all notebooks and debugger and Open new Debugger without active notebook.


KsavinN:
![restoreIssue](https://user-images.githubusercontent.com/34033913/69613518-f8183780-1031-11ea-85bd-1504b307e245.gif)

jtpio:
Looking into that one. It looks like there is a dangling reference to the last session.

Issue #168, opened 2019-11-12
Breakpoints not removed from the sidebar on notebook closed


jtpio:
When a notebook is closed, the breakpoints stay in the sidebar and are not removed.

Clicking on "Remove All Breakpoints" throws the following error:

```
Uncaught (in promise) Error: A kernel is required to send debug messages.
```

![breakpoints-notebook-closed](https://user-images.githubusercontent.com/591645/68663832-9ccd4c00-053f-11ea-9f7a-33aa08f1c97f.gif)

jtpio:
Looking into this. There might be some similarities with https://github.com/jupyterlab/debugger/issues/181.

As a side note, the "Remove All Breakpoints" button could be disabled when there is no breakpoint, or when there is no active debug session.

jtpio:
The state of the sidebar will normally be handled by the toggleable button: https://github.com/jupyterlab/debugger/issues/214

PR #234, opened 2019-11-26
Add plugin to track notebooks, consoles and editors

jtpio:
~This PR builds on top of #231 should be rebased when #231 is merged.~ [done]

This change lays the foundations for #176, #212 and #224.

This idea is have a plugin that depends on `INotebookTracker`, `IConsoleTracker` and `IEditorTracker` (all optional), that will be used to find the `CodeEditor.IEditor` from where an event originated. An event could be:

- a `stopped` event that contains the source path
- a click on a breakpoint (#224) (which also contains the source path)

The `TrackerHandler` iterates through all the widgets of the trackers to try to find a matching hash. When there is no match, this is where we could extend the logic to open files in the main area (#212) or to the read-only code editor view.

### DIFF
--- a/src/handlers/tracker.ts
+++ b/src/handlers/tracker.ts
@@ -16,7 +16,7 @@ import { Debugger } from '../debugger';
 
 import { IDebugger } from '../tokens';
 
-export class WidgetHandler implements IDisposable {
+export class TrackerHandler implements IDisposable {
   constructor(options: DebuggerWidgetHandler.IOptions) {
     this.debuggerService = options.debuggerService;
     this.notebookTracker = options.notebookTracker;
--- a/src/handlers/tracker.ts
+++ b/src/handlers/tracker.ts
@@ -10,10 +10,11 @@ import { Signal } from '@phosphor/signaling';
 
 import { Callstack } from '../callstack';
 
+import { EditorHandler } from './editor';
+
 import { Debugger } from '../debugger';
 
 import { IDebugger } from '../tokens';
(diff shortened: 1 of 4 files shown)


src/handlers/tracker.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  protected onCurrentFrameChanged(
+    callstackModel: Callstack.Model,
+    frame: Callstack.IFrame
+  ) {
+    const debugSessionPath = this.debuggerService.session.client.path;
+    this.handleNotebooks(debugSessionPath, frame);
+    this.handleConsoles(debugSessionPath, frame);
+    this.handleEditors(debugSessionPath, frame);
```
jtpio: These methods could be renamed `findInNotebooks`, `findInConsoles` and `findInEditors` (or something similar).

jtpio: Done.

jtpio:
This should also make it easier to sync the visual elements across widgets with the same session paths. For example:

editor-tracker

jtpio:
This is ready for a first look.

cc @JohanMabille this is what we discussed in-person last week. Intent and summary in the [first post](https://github.com/jupyterlab/debugger/pull/234#issue-345867159).

jtpio:
This could also help fixing https://github.com/jupyterlab/debugger/issues/168 and https://github.com/jupyterlab/debugger/issues/230.

jtpio:
@KsavinN I'll merge this one after the tests pass.

PR #237, opened 2019-11-28
IDebugger is not an IDisposable anymore

### LINKED ISSUE #235: Should IDebugger be an IDisposable?
`IDebugger` is the public token that other plugins can import. Is there a use case for a consumer of `IDebugger` having the ability to call `IDebugger#dispose()`? It seems like that should not be part of the public interface.

JohanMabille:
Fixes #235

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -3,6 +3,8 @@
 
 import { IClientSession } from '@jupyterlab/apputils';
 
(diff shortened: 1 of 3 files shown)


afshin:
The service might still reasonably need to be `IDisposable`. I would suggest *only* removing `IDisposable` from the public token's interface. (That way the tests remain the same as well.)

Issue #154, opened 2019-11-05
Create a function for adding toolbar buttons.


afshin:
Create a helper function that adds the default debugger launch command to the toolbar of a `ConsolePanel` or `NotebookPanel` *only if its kernel supports debugging*. This function should be called from within the `consoles` plugin and the `notebooks` plugin.

```typescript
/**
 * Adds a toolbar button to open the debugger.
 */
function updateToolbar(widget: MainAreaWidget<ConsolePanel | NotebookPanel>): void {
  /* ... */
}
```

jtpio:
This was addressed in #239.

PR #239, opened 2019-11-29
Add life-cycle button in toolbar of notebook and console panel.

KsavinN:
related to #214  and #154 

Now: 

- [x] toggle should stop/start session.

- [x] toggle should appear only for the supported kernel.

- [x] user cant run debugging/ set breakpoints it toggle is off

- [x] if user set breakpoints and reload page client will be automatically turn on the toggle

### DIFF
--- a/src/handlers/console.ts
+++ b/src/handlers/console.ts
@@ -47,6 +47,7 @@ export class ConsoleHandler implements IDisposable {
     this._cellMap.values().forEach(handler => handler.dispose());
     this._cellMap.dispose();
     Signal.clearData(this);
+    this.consolePanel.node.setAttribute('data-jp-debugger', 'false');
   }
 
   protected addEditorHandler(cell: Cell) {
--- a/src/handlers/console.ts
+++ b/src/handlers/console.ts
@@ -47,7 +47,6 @@ export class ConsoleHandler implements IDisposable {
     this._cellMap.values().forEach(handler => handler.dispose());
     this._cellMap.dispose();
     Signal.clearData(this);
-    this.consolePanel.node.removeAttribute('data-jp-debugger');
   }
 
   protected addEditorHandler(cell: Cell) {
--- a/src/handlers/console.ts
+++ b/src/handlers/console.ts
@@ -47,7 +47,7 @@ export class ConsoleHandler implements IDisposable {
     this._cellMap.values().forEach(handler => handler.dispose());
     this._cellMap.dispose();
     Signal.clearData(this);
-    this.consolePanel.node.setAttribute('data-jp-debugger', 'false');
+    this.consolePanel.node.removeAttribute('data-jp-debugger');
   }
 
   protected addEditorHandler(cell: Cell) {
--- a/src/handlers/notebook.ts
+++ b/src/handlers/notebook.ts
@@ -37,7 +37,6 @@ export class NotebookHandler implements IDisposable {
     this._cellMap.values().forEach(handler => handler.dispose());
     this._cellMap.dispose();
     Signal.clearData(this);
-    this.notebookPanel.node.removeAttribute('data-jp-debugger');
   }
 
   protected addEditorHandler(cell: Cell) {
--- a/src/handlers/notebook.ts
+++ b/src/handlers/notebook.ts
@@ -37,6 +37,7 @@ export class NotebookHandler implements IDisposable {
     this._cellMap.values().forEach(handler => handler.dispose());
     this._cellMap.dispose();
     Signal.clearData(this);
+    this.notebookPanel.node.setAttribute('data-jp-debugger', 'false');
   }
 
   protected addEditorHandler(cell: Cell) {
--- a/src/handlers/notebook.ts
+++ b/src/handlers/notebook.ts
@@ -37,7 +37,7 @@ export class NotebookHandler implements IDisposable {
     this._cellMap.values().forEach(handler => handler.dispose());
     this._cellMap.dispose();
     Signal.clearData(this);
-    this.notebookPanel.node.setAttribute('data-jp-debugger', 'false');
+    this.notebookPanel.node.removeAttribute('data-jp-debugger');
   }
 
   protected addEditorHandler(cell: Cell) {
--- a/src/handlers/tracker.ts
+++ b/src/handlers/tracker.ts
@@ -58,19 +58,14 @@ export class TrackerHandler implements IDisposable {
     this.consoleTracker = options.consoleTracker;
     this.editorTracker = options.editorTracker;
 
-    this.debuggerService.modelChanged.connect(() => {
-      const debuggerModel = this.debuggerService.model as Debugger.Model;
-      if (debuggerModel) {
-        debuggerModel.callstack.currentFrameChanged.connect(
-          this.onCurrentFrameChanged,
-          this
-        );
+    this.readOnlyEditorFactory = new ReadOnlyEditorFactory({
+      editorServices: options.editorServices
+    });
 
-        debuggerModel.breakpoints.clicked.connect((_, breakpoint) => {
-          const debugSessionPath = this.debuggerService.session.client.path;
-          this.find(debugSessionPath, breakpoint.source.path);
-        });
-      }
+    this.readOnlyEditorTracker = new WidgetTracker<
+      MainAreaWidget<CodeEditorWrapper>
+    >({
+      namespace: '@jupyterlab/debugger'
     });
 
     this.onModelChanged();
@@ -258,7 +253,7 @@ export class TrackerHandler implements IDisposable {
 
   protected findInReadOnlyEditors(_: string, source: string) {
     let editors: CodeEditor.IEditor[] = [];
-    this.readOnlyEditorTracker?.forEach(widget => {
+    this.readOnlyEditorTracker.forEach(widget => {
       const editor = widget.content?.editor;
       if (!editor) {
         return;
--- a/src/handlers/tracker.ts
+++ b/src/handlers/tracker.ts
@@ -58,14 +58,19 @@ export class TrackerHandler implements IDisposable {
     this.consoleTracker = options.consoleTracker;
     this.editorTracker = options.editorTracker;
 
-    this.readOnlyEditorFactory = new ReadOnlyEditorFactory({
-      editorServices: options.editorServices
-    });
+    this.debuggerService.modelChanged.connect(() => {
+      const debuggerModel = this.debuggerService.model as Debugger.Model;
+      if (debuggerModel) {
+        debuggerModel.callstack.currentFrameChanged.connect(
+          this.onCurrentFrameChanged,
+          this
+        );
 
-    this.readOnlyEditorTracker = new WidgetTracker<
-      MainAreaWidget<CodeEditorWrapper>
-    >({
-      namespace: '@jupyterlab/debugger'
+        debuggerModel.breakpoints.clicked.connect((_, breakpoint) => {
+          const debugSessionPath = this.debuggerService.session.client.path;
+          this.find(debugSessionPath, breakpoint.source.path);
+        });
+      }
     });
 
     this.onModelChanged();
@@ -253,7 +258,7 @@ export class TrackerHandler implements IDisposable {
 
   protected findInReadOnlyEditors(_: string, source: string) {
     let editors: CodeEditor.IEditor[] = [];
-    this.readOnlyEditorTracker.forEach(widget => {
+    this.readOnlyEditorTracker?.forEach(widget => {
       const editor = widget.content?.editor;
       if (!editor) {
         return;
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,12 @@ import {
   JupyterFrontEndPlugin
 } from '@jupyterlab/application';
 
-import { IClientSession, ICommandPalette } from '@jupyterlab/apputils';
+import {
+  IClientSession,
+  ICommandPalette,
+  Toolbar,
+  ToolbarButton
+} from '@jupyterlab/apputils';
 
 import { IEditorServices } from '@jupyterlab/codeeditor';
 
@@ -83,6 +88,47 @@ async function setDebugSession(
   app.commands.notifyCommandChanged();
 }
 
+function updateToolbar(
+  widget: NotebookPanel | ConsolePanel | FileEditor
+): void {
+  console.log({ widget });
+
+  const isConsolePanel = widget instanceof ConsolePanel;
+  let toogleState = false;
+  const getToolbar = (): Toolbar => {
+    if (isConsolePanel) {
+      return (
+        ((widget as ConsolePanel).widgets.find(
+          widget => widget instanceof Toolbar
+        ) as Toolbar) || new Toolbar()
+      );
+    } else {
+      return (widget as NotebookPanel)?.toolbar;
+    }
+  };
+
+  const insertItemAndCheckIfExist = (toolbar: Toolbar) => {
+    return toolbar.addItem(
+      'debugger-lifeCycle-button',
+      new ToolbarButton({
+        className: 'jp-debugger-switch-button',
+        iconClassName: 'jp-toggle-switch',
+        onClick: () => {
+          toogleState = !toogleState;
+          widget.node.setAttribute('data-debugger-on', toogleState.toString());
+        },
+        tooltip: 'Enable / Disable Debugger'
+      })
+    );
+  };
+
+  const toolbar = getToolbar();
+
+  if (insertItemAndCheckIfExist(toolbar) && isConsolePanel) {
+    (widget as ConsolePanel).insertWidget(0, toolbar);
+  }
+}
+
 class DebuggerHandler<
   H extends ConsoleHandler | NotebookHandler | FileHandler
 > {
@@ -98,6 +144,7 @@ class DebuggerHandler<
     if (!debug.model || this.handlers[widget.id] || !debuggingEnabled) {
       return;
     }
+    updateToolbar(widget);
     const handler = new this.builder({
       debuggerService: debug,
       widget
@@ -108,7 +155,6 @@ class DebuggerHandler<
       handler.dispose();
       delete this.handlers[widget.id];
     });
-
     debug.model.disposed.connect(async () => {
       const handlerIds = Object.keys(this.handlers);
       if (handlerIds.length === 0) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -33,6 +33,8 @@ import { Debugger } from './debugger';
 
 import { ConsoleHandler } from './handlers/console';
 
+import { CommandRegistry } from '@phosphor/commands';
+
 import { FileHandler } from './handlers/file';
 
 import { NotebookHandler } from './handlers/notebook';
@@ -89,12 +91,17 @@ async function setDebugSession(
 }
 
 function updateToolbar(
-  widget: NotebookPanel | ConsolePanel | FileEditor
+  widget: NotebookPanel | ConsolePanel | FileEditor,
+  debug: IDebugger,
+  commands: CommandRegistry,
+  handler: any
 ): void {
-  console.log({ widget });
-
   const isConsolePanel = widget instanceof ConsolePanel;
-  let toogleState = false;
+  const checkState = () =>
+    debug.session.client.name === debug.session.currentStateClient;
+
+  widget.node.setAttribute('data-debugger-on', checkState().toString());
+
   const getToolbar = (): Toolbar => {
     if (isConsolePanel) {
       return (
@@ -113,9 +120,17 @@ function updateToolbar(
       new ToolbarButton({
         className: 'jp-debugger-switch-button',
         iconClassName: 'jp-toggle-switch',
-        onClick: () => {
-          toogleState = !toogleState;
-          widget.node.setAttribute('data-debugger-on', toogleState.toString());
+        onClick: async () => {
+          if (debug.model == null && !checkState()) {
+            await commands.execute(CommandIDs.create);
+            handler.update(debug, widget);
+          }
+          if (!checkState()) {
+            debug.session.currentStateClient = debug.session.client.name;
+          } else {
+            debug.session.currentStateClient = null;
+          }
+          widget.node.setAttribute('data-debugger-on', checkState().toString());
         },
         tooltip: 'Enable / Disable Debugger'
       })
@@ -144,13 +159,14 @@ class DebuggerHandler<
     if (!debug.model || this.handlers[widget.id] || !debuggingEnabled) {
       return;
     }
-    updateToolbar(widget);
+
     const handler = new this.builder({
       debuggerService: debug,
       widget
     });
     widget.node.setAttribute('data-jp-debugger', 'true');
     this.handlers[widget.id] = handler;
+
     widget.disposed.connect(() => {
       handler.dispose();
       delete this.handlers[widget.id];
@@ -260,8 +276,10 @@ const notebooks: JupyterFrontEndPlugin<void> = {
       if (!(widget instanceof NotebookPanel)) {
         return;
       }
+
       await setDebugSession(app, debug, widget.session);
-      void handler.update(debug, widget);
+      handler.update(debug, widget);
+      updateToolbar(widget, debug, app.commands, handler);
     });
   }
 };
--- a/src/index.ts
+++ b/src/index.ts
@@ -151,13 +151,15 @@ class DebuggerHandler<
     widget: W
   ): Promise<void> {
     const debuggingEnabled = await debug.requestDebuggingEnabled();
-    if (
-      !debug.model ||
-      this.handlers[widget.id] ||
-      !debuggingEnabled ||
-      !debug.isStarted
-    ) {
+    if (!debug.model || !debuggingEnabled || !debug.isStarted) {
       return;
+    }
+
+    if (
+      this.handlers[widget.id] &&
+      !debug.session.debuggedClients.has(debug.session.client.path)
+    ) {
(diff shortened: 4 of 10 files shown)


KsavinN:
onlyUI

Implement simple button switch in the toolbar of notebook and insert toolbar with button switch to consolepanel

src/handlers/notebook.ts
```diff
@@ -23,6 +24,22 @@ export class NotebookHandler implements IDisposable {
     this.debuggerService = options.debuggerService;
     this.notebookPanel = options.widget;
 
+    this.notebookPanel.toolbar.addItem(
```
jtpio: Wondering whether we could use the `insertAfter` method from the toolbar, so it would look something like this:

```typescript
this.notebookPanel.toolbar.insertAfter('kernelStatus', 'debugger-lifeCycle-button', new ToolbarButton(...));`
```

jtpio:
Thanks for posting a screencast, this looks promising!

src/handlers/console.ts
```diff
@@ -32,6 +33,24 @@ export class ConsoleHandler implements IDisposable {
...
+        className: 'jp-debugger-switch-button',
+        iconClassName: 'jp-toggle-switch',
+        onClick: () => {
+          this.toogleState = !this.toogleState;
+          this.consolePanel.node.setAttribute(
+            'data-debugger-on',
+            this.toogleState.toString()
+          );
+        },
+        tooltip: 'ON/OFF Debugger state'
```
jtpio: Maybe something like "Enable / Disable Debugger" would be more user-friendly?

src/handlers/notebook.ts
```diff
@@ -15,13 +15,30 @@ import { Debugger } from '../debugger';
...
 
+import { ToolbarButton } from '@jupyterlab/apputils';
 import { IDebugger } from '../tokens';
 
 export class NotebookHandler implements IDisposable {
   constructor(options: NotebookHandler.IOptions) {
     this.debuggerModel = options.debuggerService.model as Debugger.Model;
     this.debuggerService = options.debuggerService;
     this.notebookPanel = options.widget;
+    this.notebookPanel.toolbar.insertAfter(
```
jtpio: The code to add a new button is almost the same as for the console (and later for the editors too).

There is an issue to create a function for adding debug buttons in https://github.com/jupyterlab/debugger/issues/154. Should we refactor that in this PR or in a separate change?

KsavinN: Good catch. I will try to refactor this PR.

src/handlers/console.ts
KsavinN: You definitely, absolutely have right.

src/index.ts
```diff
@@ -90,6 +92,58 @@ async function setDebugSession(
   app.commands.notifyCommandChanged();
 }
 
+function updateToolbar(
+  widget: NotebookPanel | ConsolePanel | FileEditor,
+  debug: IDebugger,
+  app: JupyterFrontEnd
```
afshin: Instead of passing the whole application, I would suggest just passing the `CommandRegistry` that executes the command you need, because you're not using any other app functionality.

src/service.ts
```diff
@@ -150,6 +150,13 @@ export class DebugService implements IDebugger, IDisposable {
     Signal.clearData(this);
   }
 
+  /**
+   * is actual kernel state
```
afshin: This function's name is a little confusing. I'm not sure what it should be called, but I think it should have a more descriptive name.

src/index.ts
```diff
@@ -90,6 +94,60 @@ async function setDebugSession(
...
+
+  const getToolbar = (): Toolbar => {
+    if (isConsolePanel) {
+      return (
+        ((widget as ConsolePanel).widgets.find(
+          widget => widget instanceof Toolbar
+        ) as Toolbar) || new Toolbar()
+      );
+    } else {
+      return (widget as NotebookPanel)?.toolbar;
```
jtpio: If the `?` is necessary here, should we also have it above? (for example for `widget as ConsolePanel`)

jtpio:
It would be nice to rebase this PR to grab all the latest changes from `master` (including the single mode debugger).

jtpio:
Also @KsavinN let me know if you want to share the work on this. Otherwise I'm also happy to pick this up :)

KsavinN:
> Also @KsavinN let me know if you want to share the work on this. Otherwise I'm also happy to pick this up :)

I start rebasing. And i think i will need today spend some time for reading all your changes which you made :)

src/handlers/console.ts
```diff
@@ -39,6 +39,7 @@ export class ConsoleHandler implements IDisposable {
     this.isDisposed = true;
     this.editorHandler.dispose();
     Signal.clearData(this);
+    this.consolePanel.node.setAttribute('data-jp-debugger', 'false');
```
afshin: This should be `removeAttribute` I think.

src/handlers/tracker.ts
```diff
@@ -58,14 +58,19 @@ export class TrackerHandler implements IDisposable {
...
+        debuggerModel.callstack.currentFrameChanged.connect(
+          this.onCurrentFrameChanged,
+          this
+        );
+
+        debuggerModel.breakpoints.clicked.connect((_, breakpoint) => {
+          const debugSessionPath = this.debuggerService.session.client.path;
+          this.find(debugSessionPath, breakpoint.source.path);
+        });
+      }
```
jtpio: This snippet of code is already in the `onModelChanged` below. Is there a reason to have it here too?

src/handlers/tracker.ts
```diff
@@ -253,7 +258,7 @@ export class TrackerHandler implements IDisposable {
 
   protected findInReadOnlyEditors(_: string, source: string) {
     let editors: CodeEditor.IEditor[] = [];
-    this.readOnlyEditorTracker.forEach(widget => {
+    this.readOnlyEditorTracker?.forEach(widget => {
```
jtpio: Is `?` needed? `readOnlyEditorTracker` is instantiated in the constructor and should not be `null` or `undefined`.

src/session.ts
```diff
@@ -60,13 +60,16 @@ export class DebugSession implements IDebugger.ISession {
       this._client.iopubMessage.disconnect(this._handleEvent, this);
     }
 
-    this._client = client;
-
     if (client) {
+      this._client = client;
```
jtpio: Should this line be moved? I think it was left outside the `if` to be able to set the client to `null`.

src/session.ts
```diff
@@ -60,13 +60,16 @@ export class DebugSession implements IDebugger.ISession {
...
       this._client.iopubMessage.connect(this._handleEvent, this);
     }
   }
 
+  get debuggedClients(): Set<string> {
```
jtpio: Does this mean that a debug session should keep track of multiple clients?

jtpio:
Just tried the change locally. This button looks like a good way to enable and disable the debugger :+1: 

It seems we might have "lost" the breakpoint marker when debugging files and hovering on the gutter:

debug-files-hover

src/handlers/tracker.ts
KsavinN: i had problem during unit test that's why i put that there. I will try without ?

src/session.ts
KsavinN: this was only my suggestion. For now, we still tracking all clients which are have supported kernel.

src/handlers/tracker.ts
afshin: Yeah, it looks like the lines below this chunk need to be removed?

src/index.ts
```diff
@@ -79,31 +84,95 @@ async function setDebugSession(
...
-  update<W extends ConsolePanel | NotebookPanel | FileEditor>(
+  async update<W extends ConsolePanel | NotebookPanel | FileEditor>(
     debug: IDebugger,
     widget: W
-  ): void {
+  ): Promise<void> {
     if (!debug.model || this.handlers[widget.id] || !debug.isDebuggingEnabled) {
+      if (!debug.session.debuggedClients.has(debug.session.client.name)) {
+        // in setDebugSession allways we start session to get codeId, if debugging toggle is off we now stop session
+        await debug.stop();
```
afshin: This could just be `return debug.stop();` instead of awaiting and returning below.

afshin: (And you you as well remove the comment.)

src/index.ts
```diff
@@ -116,6 +185,11 @@ class DebuggerHandler<
       });
       this.handlers = {};
     });
+
+    if (!debug.session.debuggedClients.has(debug.session.client.name)) {
+      // in setDebugSession allways we start session to get codeId, if debugging toggle is off we now stop session
+      await debug.stop();
```
afshin: Same as above.

src/index.ts
```diff
@@ -79,31 +84,95 @@ async function setDebugSession(
...
+
+  const changeAtribute = () => {
+    if (debug.session.debuggedClients.has(debug.session.client.name)) {
+      widget.node.setAttribute('data-jp-debugger', 'true');
+    } else {
+      widget.node.removeAttribute('data-jp-debugger');
+    }
+  };
+
+  const getToolbar = (): Toolbar => {
```
afshin: Hm, aren't the three types of things that we support all main area widgets? They should all have a `.content` and a `.toolbar` so you shouldn't have to write a function to get the toolbar yourself.

KsavinN: Only Console widget don't have toolbar and content. DocumentWidget and NotebookPanel have a toolbar. That's why I need a function for check if toolbar exists.

console

src/handlers/console.ts
```diff
@@ -47,6 +47,7 @@ export class ConsoleHandler implements IDisposable {
     this._cellMap.values().forEach(handler => handler.dispose());
     this._cellMap.dispose();
     Signal.clearData(this);
+    this.consolePanel.node.removeAttribute('data-jp-debugger');
```
jtpio: It looks a bit odd for this attribute to be set at the plugin level, and then removed here on dispose in the handler.

Maybe adding and removing the attribute should happen at the same place?

src/handlers/tracker.ts
jtpio: What problem?

src/service.ts
```diff
@@ -221,20 +228,27 @@ export class DebugService implements IDebugger, IDisposable {
...
+    if (!this.isStarted() && autoStart) {
       await this.start();
     }
 
-    if (stoppedThreads.size !== 0) {
-      await this.getAllFrames();
-    } else {
-      this.clearModel();
-      this.clearSignals();
+    if (this.isLifeCycleOnKernelState()) {
```
jtpio: How about using an early return instead?

```typescript
if (!this.isLifeCycleOnKernelState()) {
   return;
}
```

So it's more readable, and the rest of the code won't even show in the diff anymore.

style/icons.css
```diff
@@ -86,3 +86,46 @@
 [data-jp-theme-light='false'] .jp-ViewBreakpointIcon {
   background-image: url('icons/md-dark/view-breakpoint.svg');
 }
+
+.jp-debugger-switch-button::before {
```
jtpio: Since those are new classes, should we try to follow the same CSS naming convention?

style/icons.css
```diff
@@ -86,3 +86,46 @@
...
+  margin-left: 5px;
+  margin-right: 50%;
+  background-color: white;
+  -webkit-transition: 0.4s;
+  transition: 0.4s;
+  border-radius: 50%;
+}
+
+[data-jp-debugger='true'] .jp-toggle-switch {
+  background-color: #f17c00;
```
jtpio: Maybe we could use a lab variable here? Either the `--jp-warn-color*` or `--md-orange-***`

src/handlers/notebook.ts
jtpio: Let's not do that for now, since the button is also added to other toolbars than the notebook.

src/index.ts
```diff
@@ -79,31 +84,94 @@ async function setDebugSession(
...
+        ((widget as ConsolePanel).widgets.find(
+          widget => widget instanceof Toolbar
+        ) as Toolbar) || new Toolbar()
+      );
+    } else {
+      return (widget as NotebookPanel | DocumentWidget)?.toolbar;
+    }
+  };
+
+  const insertItemAndCheckIfExist = (toolbar: Toolbar) => {
```
jtpio: Do we need a function that does 2 things at the same time, if it's only used once below?

src/index.ts
afshin: Hm, this seems like a shortcoming in JupyterLab. I think the console should probably be a main area widget as well, for exactly this reason.

src/handlers/tracker.ts
KsavinN: issue

jtpio: Well, the diff for this PR shows that it is not instantiated anymore, so that's probably why:

image

KsavinN: Thanks, @jtpio . Probably I removed that during rebasing

src/index.ts
afshin: cf. https://github.com/jupyterlab/jupyterlab/issues/7624

src/index.ts
```diff
@@ -79,31 +84,91 @@ async function setDebugSession(
...
+        ) as Toolbar) || new Toolbar()
+      );
+    } else {
+      return (widget as NotebookPanel | DocumentWidget)?.toolbar;
+    }
+  };
+
+  const toolbar = getToolbar();
+
+  const isToolbarExist = toolbar.addItem(
```
jtpio: This name sounds a bit misleading. `addItem` returns `true` if the item was added to the toolbar, `false` if it's already there.

src/index.ts
```diff
@@ -79,31 +84,91 @@ async function setDebugSession(
...
+        changeAtribute();
+      },
+      tooltip: 'Enable / Disable Debugger'
+    })
+  );
+
+  if (isToolbarExist && isConsolePanel) {
+    (widget as ConsolePanel).insertWidget(0, toolbar);
+  }
+  changeAtribute();
```
jtpio: How about this?

```suggestion
  toggleAttribute();
```

src/index.ts
```diff
@@ -79,31 +84,91 @@ async function setDebugSession(
...
+    if (isConsolePanel) {
+      return (
+        ((widget as ConsolePanel).widgets.find(
+          widget => widget instanceof Toolbar
+        ) as Toolbar) || new Toolbar()
+      );
+    } else {
+      return (widget as NotebookPanel | DocumentWidget)?.toolbar;
+    }
+  };
```
jtpio: Could something like this be more readable? Also it makes it more apparent that `ConsolePanel` is (for now) a special case:

```typescript
  const getToolbar = (): Toolbar => {
    if (!(widget instanceof ConsolePanel)) {
      return widget.toolbar;
    }
    const toolbar = widget.widgets.find(w => w instanceof Toolbar) as Toolbar;
    return toolbar ?? new Toolbar();
  };
```

src/service.ts
```diff
@@ -129,6 +129,13 @@ export class DebugService implements IDebugger, IDisposable {
     Signal.clearData(this);
   }
 
+  /**
+   * Checking lifecycle state of debugger of current client
+   */
+  isLifeCycleOnKernelState(): boolean {
```
jtpio: This could be made `private`.

jtpio: Or maybe not even necessary if used only once within the service?

KsavinN: I thought we will use that in another place. For now in service we only used it in one place. It's better to remove that for now

jtpio:
This needs a rebase to take #275 into account.

src/index.ts
```diff
@@ -79,31 +84,86 @@ async function setDebugSession(
...
 
+function updateToolbar(
+  widget: NotebookPanel | ConsolePanel | DocumentWidget,
+  debug: IDebugger
+): void {
+  if (!debug.isDebuggingEnabled) {
+    return;
+  }
+
+  const toggleAtribute = () => {
```
jtpio: Typo:

```suggestion
  const toggleAttribute = () => {
```

src/index.ts
```diff
@@ -93,22 +149,27 @@ class DebuggerHandler<
...
     const handler = new this.builder({
       debuggerService: debug,
       widget
     });
-    widget.node.setAttribute('data-jp-debugger', 'true');
+
     this.handlers[widget.id] = handler;
+
     widget.disposed.connect(() => {
+      widget.node.removeAttribute('data-jp-debugger');
```
jtpio: Is this line required? (since the widget is disposed)

src/index.ts
```diff
@@ -93,22 +149,27 @@ class DebuggerHandler<
   async update<W extends ConsolePanel | NotebookPanel | FileEditor>(
     debug: IDebugger,
     widget: W
-  ) {
+  ): Promise<void> {
     const debuggingEnabled = await debug.requestDebuggingEnabled();
     if (!debug.model || this.handlers[widget.id] || !debuggingEnabled) {
+      if (!debug.session.debuggedClients.has(debug.session.client.name)) {
+        return debug.stop();
```
jtpio: This means that `debug.stop()` is called for clients that do not support debugging, which brings back #268 (fixed in #275)

jtpio: This is visible in the notebook server logs:

```
[IPKernelApp] ERROR | UNKNOWN CONTROL MESSAGE TYPE: 'debug_request'
```

src/index.ts
```diff
@@ -122,6 +183,10 @@ class DebuggerHandler<
       });
       this.handlers = {};
     });
+
+    if (!debug.session.debuggedClients.has(debug.session.client.name)) {
```
jtpio: Should this be `debug.session.client.path` instead?

jtpio: (same question for all other places where it's used)

src/index.ts
jtpio: Maybe it should be moved in the `debug.model.disposed` below?

jtpio:
Things to follow-up on in separate PRs:

- Should the sidebar always be shown once we explicitly started the debugger? Otherwise should it be automatically created when we switch the button on?
- Should the toggle buttons be shown only if the sidebar is visible?
- Closing the sidebar should remove the buttons (or remove them).

jasongrout:
I'm confused - what *exactly* is that toggle doing? Is it equivalent to starting a kernel and shutting down the kernel? Or is it something more specific about debugging?

jasongrout:
Ah, reading the code, I see it is about toggling a debugger session, not a kernel session.

Is there any reason we'd ever turn it off? For example, in VS Code, I can *always* set a breakpoint, there is no modality in the debugger.

afshin:
@jasongrout Yes, you might want *one* of your notebooks to have debugging turned on while a different notebook does not have debugging turned on. In this case, when you switch notebooks, the content of the debugger will not change, so you will be able to refer to, for example, the variable state of the first notebook.

At first we considered just having *all* debug-able things automatically start debugging, but I think having a toggle facilitates more workflows. (And a big toggle is pretty discoverable.)

src/index.ts
```diff
@@ -93,22 +149,28 @@ class DebuggerHandler<
...
     widget: W
-  ) {
+  ): Promise<void> {
     const debuggingEnabled = await debug.requestDebuggingEnabled();
-    if (!debug.model || this.handlers[widget.id] || !debuggingEnabled) {
+    if (
+      !debug.model ||
+      this.handlers[widget.id] ||
+      !debuggingEnabled ||
+      !debug.isStarted()
```
KsavinN: I added  !debug.isStarted() due to this error 
issue2

before update or create handler always is stared before setDebugSession where we restore and explicitly set debug session starts.

jasongrout:
> when you switch notebooks, the content of the debugger will not change, so you will be able to refer to, for example, the variable state of the first notebook.

So a debugger is hooked up to whatever the current notebook is (that has debugging on), and you don't have the notion of having different debuggers hooked up to different kernels? Can you not debug two notebooks simultaneously?

afshin:
> So a debugger is hooked up to whatever the current notebook is (that has debugging on), and you don't have the notion of having different debuggers hooked up to different kernels? Can you not debug two notebooks simultaneously?

You can debug two notebooks simultaneously if you like, *especially* if you're just using the debugger as a variable inspector. But there is only one debugger UI, so only the most recently focused notebook (which has debugging toggled on) will be shown in the debugger UI.

This means that if you *want* to see the variables from `foo.ipynb` while you are inside `bar.ipynb` you can. But also if you want to just see `foo.ipynb` when it is focused and `bar.ipynb` when it is focused, you can do that as well.

jasongrout:
If I have nbA and nbB, both with debugging turned on, and I want to look at nbA while seeing the debug window for nbB, I turn off debugging for nbA. Does that mess up the debugging session for nbA? Can I turn it back on and continue debugging nbA, with all my breakpoints set, etc.? In other words, is the switch mainly about toggling the single debug UI, or is it deeper in that I have to restart debugging nbA?

SylvainCorlay:
> and I want to look at nbA while seeing the debug window for nbB.

I don't think this is really possible right now.

jtpio:
> If I have nbA and nbB, both with debugging turned on, and I want to look at nbA while seeing the debug window for nbB, I turn off debugging for nbA. Does that mess up the debugging session for nbA?

The simpler option here would be to focus on nbB to look at the debug window for nbB, and focus back on nbA to look at the debug window for nbA?

Switching off debugging (via the button in this PR) will stop the debugger, and the state of the debugging session will be lost.

Here is a screencast from a while back to illustrate the flow:

multi-notebook-debugging

Do you think there is a use case for looking at the debug state for a notebook that doesn't have the focus?

src/index.ts
jtpio: Thanks @afshin for opening this issue.

For now we can use the current workaround, and open an issue in this repo to update the code later.

SylvainCorlay:
> Do you think there is a use case for looking at the debug state for a notebook that doesn't have the focus?

I think that it would confuse the user experience.

SylvainCorlay:
That makes me realize it may be important to indicate which kernel is being debugged in the debugger pane.

afshin:
> @jtpio said: Do you think there is a use case for looking at the debug state for a notebook that doesn't have the focus?

> @SylvainCorlay said: I think that it would confuse the user experience.

This came out of interviews I did with some potential users who specifically said they would like to have a view on the variable state of **Notebook A** potentially one in a breakpoint-stopped state while they were playing around with ideas in **Notebook B**. The toggle was to allow for both default-on (i.e., all notebooks always have debugging) and also for occasionally turning one off so you can look at a previous notebook's state.

jtpio:
> That makes me realize it may be important to indicate which kernel is being debugged in the debugger pane.

Yes, we discussed that last time with @afshin and @tgeorgeux and it sounds like a good idea (but let's track that in a different issue).

SylvainCorlay:
>>@jtpio said: Do you think there is a use case for looking at the debug state for a notebook that doesn't have the focus?
>
> >@SylvainCorlay said: I think that it would confuse the user experience.
>
> @afshin This came out of interviews I did with some potential users who specifically said they would like to have a view on the variable state of Notebook A potentially one in a breakpoint-stopped state while they were playing around with ideas in Notebook B. The toggle was to allow for both default-on (i.e., all notebooks always have debugging) and also for occasionally turning one off so you can look at a previous notebook's state.

Do you think we could do this after 0.1?

> @jtpio Yes, we discussed that last time with @afshin and @tgeorgeux and it sounds like a good idea (but let's track that in a different issue).

Sounds good

afshin:
> Do you think we could do this after 0.1?

The toggle behavior in this PR already supports that. But we should review workflows more carefully after 0.1, yes.

src/index.ts
jtpio: Was this reverted in the last rebase?

jtpio: OK the code was restructured slightly it seems.

src/index.ts
```diff
@@ -93,22 +149,30 @@ class DebuggerHandler<
...
-  ) {
+  ): Promise<void> {
     const debuggingEnabled = await debug.requestDebuggingEnabled();
-    if (!debug.model || this.handlers[widget.id] || !debuggingEnabled) {
+    if (!debug.model || !debuggingEnabled || !debug.isStarted()) {
+      return;
+    }
+
+    if (this.handlers[widget.id]) {
+      if (!debug.session.debuggedClients.has(debug.session.client.path)) {
```
jtpio: Maybe the two `if` could be combined then?

SylvainCorlay:
> The toggle behavior in this PR already supports that. But we should review workflows more carefully after 0.1, yes.

I thought it would simply make it behave like a non-debugging kernel?

src/session.ts
jtpio: Not sure yet the `session` is the preferred place to store such information.

But it would be fine to refactor that in a follow-up PR.

src/index.ts
KsavinN: yeah, definitely. Sorry need make break :)

afshin:
@SylvainCorlay What I mean is that I'm inside non-debugging **Notebook B** and I just want the debugger to still show me the debugging **Notebook A** state. This PR supports that sort of use case. But it's not more sophisticated in terms of switching kernels in the debugger (which is an idea that came out of when Tim and Jeremy and I spoke and should wait until after 0.1).

jtpio:
Since this PR has been open for a quite a while and need to be rebased regularly, I would be fine merging it as is and address the [remaining points](https://github.com/jupyterlab/debugger/pull/239#issuecomment-565490444) in follow-up changes.

@afshin what do you think?

afshin:
@jtpio Yeah, let's merge and iterate, that's better.

PR #240, opened 2019-12-02
Allow setting breakpoints in out-of-focus cells

### LINKED ISSUE #238: Allow creating and deleting breakpoints in out-of-focus cells
At the moment, adding a breakpoint to a notebook cell requires it to have focus.

I think that it would make sense to always allow for adding breakpoints to any cell.

jtpio:
Fixes #238 

Allowing setting breakpoints in other cells even when they don't have the focus does feel more natural.

Also all the gutters initialized at the same when the notebook handler is created, which also feels more intuitive. 

breakpoint-out-of-focus-cells

### DIFF
--- a/src/handlers/notebook.ts
+++ b/src/handlers/notebook.ts
@@ -1,7 +1,9 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { CodeCell } from '@jupyterlab/cells';
+import { Cell, CodeCell } from '@jupyterlab/cells';
+
+import { IObservableMap, ObservableMap } from '@jupyterlab/observables';
 
 import { Notebook, NotebookPanel } from '@jupyterlab/notebook';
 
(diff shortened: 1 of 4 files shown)


src/handlers/notebook.ts
```diff
@@ -23,9 +25,10 @@ export class NotebookHandler implements IDisposable {
     this.debuggerService = options.debuggerService;
     this.notebookPanel = options.widget;
 
-    this.createNewEditorHandler();
+    this._cellMap = new ObservableMap<EditorHandler>();
```
jtpio: This uses an `ObservableMap` instead of `Map` to follow what is used internally in the JupyerLab core notebook extension.

An `ObservableMap` is not strictly required in this case, but offers the possibility to listen on `changed` events.

jtpio:
@KsavinN I'm not sure we discussed this in a previous call or any issue other than #238. So feel free to check it out if you want.

This is ready for review.

KsavinN:
I look it locally is working well good even for split notebook widgets. Only I see is edge case for tabBars in debugger widget. But still, I think it's ready to merge.

jtpio:
Ah indeed good catch, thanks for checking.

It's probably related to the CSS change, and should be easy to fix. Will merge after.

jtpio:
This should be fixed now:

breakpoints-editors

PR #242, opened 2019-12-03
Single mode debugger

### LINKED ISSUE #176: Alternative to the main area widget tab panel

jtpio:
Fixes #176.

~Currently building on top of #241, to be rebased once it's merged.~ [done]

This implements the ideas from #176, and drops the main area widget mode for the debugger

To try it out (while this PR is open):

[Binder](https://mybinder.org/v2/gh/jtpio/debugger/single-mode?urlpath=/lab/tree/examples/index.ipynb)

### TODO

- [x] Drop `mode`
- [x] Remove the `debugger:change-mode` command
- [x] Add the `SOURCE` split to the sidebar
- [x] Open source files in the main area
- [x] Add grey background to read-only editors
- [x] Add tracker for the read-only editors to support line highlighting and revealing an editor if it is already there

single-mode-debugger2


jtpio:
This is what it currently looks like:

source-split

The logic to open the files in the main area with the button still needs to be implemented.

Also we might want to collapse the the Source split when there is nothing to show.

jtpio:
To try it out:

[Binder](https://mybinder.org/v2/gh/jtpio/debugger/single-mode?urlpath=/lab/tree/examples/index.ipynb)

SylvainCorlay:
@jtpio small nitpicking comment: a good visual indication that the SOURCE editor is read-only may be to use a gray background...

jtpio:
Good point.

We could add that to the other editors that will be added to the main area too.

SylvainCorlay:
> We could add that to the other editors that will be added to the main area too.

They would only be added with the expand button right? Should they really be read only then?

jtpio:
For the ones that correspond to cell executions that would make sense I think.

For the ones corresponding to modules (when we step into), maybe. In the case of remote kernels having them editable could lead to some confusion.

SylvainCorlay:
Really good point on remote kernels. I had not thought about it.

SylvainCorlay:
Settin g the height of the codemirror cell to 100% also looks nice in the SOURCE section:

Screenshot from 2019-12-03 18-14-42

SylvainCorlay:
Quick thought: the SOURCE section may be a more natural position for the "continue / terminate / next / step-in / step-out" buttons.

jtpio:
Looks good. And with the dark theme?

> Quick thought: the SOURCE section may be a more natural position for the "continue / terminate / next / step-in / step-out" buttons.

Started to think the same.

jtpio:
With the dark theme:

image

src/debugger.ts
```diff
@@ -88,43 +36,72 @@ export namespace Debugger {
   export class Sidebar extends SplitPanel {
     constructor(options: Sidebar.IOptions) {
       super();
+      this.id = 'jp-debugger-sidebar';
+      this.title.label = 'Environment';
       this.orientation = 'vertical';
+
+      this.addClass('jp-Debugger');
```
jtpio: We should keep only one of `jp-Debugger` or `jp-DebuggerSidebar`.

src/handlers/tracker.ts
```diff
@@ -71,6 +97,34 @@ export class TrackerHandler implements IDisposable {
...
+      mimeType,
+      path
+    });
+    const editorHandler = new EditorHandler({
+      debuggerService: this.debuggerService,
+      editor: editorWrapper.editor
+    });
+    const widget = new Panel();
+    widget.addWidget(editorWrapper);
+    widget.id = UUID.uuid4();
```
jtpio: Maybe `path` could be used for `widget.id`.

jtpio:
Update:

single-mode-debugger

jtpio:
Some things that are left but could be done in separate PRs:

- Expose the `IDebuggerReadOnlyEditorTracker` at the plugin level so it can consumed by other extensions
- Handle modules and local files: https://github.com/jupyterlab/debugger/issues/243
- Should the read only editors open in a new split?
- Center `SOURCE` editor on the current line (scroll automatically)
- Clicking on a breakpoint that points to a cell that does not exist anymore should open a new read only editor

jtpio:
This PR should now be in a usable state.

There are a couple of things that could be improved and some are listed in the [comment above](https://github.com/jupyterlab/debugger/pull/242#issuecomment-561666884).

The Binder has been rebuilt if anyone wants to test the changes.

src/handlers/tracker.ts
afshin: This can create a potential bug if the first character of the generated string is a number, which is an invalid UI in HTML. In JupyterLab, we created a helper function to guarantee this never happens: `DOMUtils.createDomID()` in the `@jupyterlab/apputils` library.

jtpio: This is indeed better, thanks.

jtpio:
@afshin did you get a chance to try this alternative to the main area widget? (locally or on [binder](https://mybinder.org/v2/gh/jtpio/debugger/single-mode?urlpath=/lab/tree/examples/index.ipynb))

SylvainCorlay:
I think this looks great and it is really well illustrated by the new UI in #247.

PR #247, opened 2019-12-05
Open the source of a breakpoint on click

### LINKED ISSUE #246: Open the source when clicking on a breakpoint
Originally posted in https://github.com/jupyterlab/debugger/pull/242#issuecomment-561666884.

If a breakpoint is in a file (temporary, local or module), and the corresponding code is not visible in the workspace, we should open a new read only editor in the main area and display its content with the breakpoints.

This will allow removing the breakpoints in a temporary file that corresponds to previous cell executions.

jtpio:
Fixes #246.

Builds on top of #242 (to be rebased).

### TODO

- [x] Activate the widget that contains the source?
- [x] Rebase

open-breakpoint-source

### DIFF
--- a/src/callstack/body.tsx
+++ b/src/callstack/body.tsx
@@ -7,8 +7,6 @@ import { Callstack } from '.';
 
 import { ReactWidget } from '@jupyterlab/apputils';
 
-import { ArrayExt } from '@phosphor/algorithm';
-
 export class Body extends ReactWidget {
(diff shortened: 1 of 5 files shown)


jtpio:
Activating the widgets makes it more natural to follow the execution and step into the code:

activate-widgets-step

jtpio:
Rebased.

PR #248, opened 2019-12-05
Add support for breakpoints in modules and files

### LINKED ISSUE #243: Adding breakpoints to local modules and files
At the moment we use the code hash when setting breakpoints. That means that we can only set breakpoints for the temporary files (`/tmp/...`) like cells. The reason it works for files is that when they are executed via the console, it is still a `request_execute`, and a new temp file is created.

Because of this, adding breakpoints to modules local to the kernels (a library for example) creates a new temporary file, and the breakpoints are ignored when stepping into the code (since the temp file is not executed).

We should look into a way to support local paths too.

jtpio:
Fixes #243.

Builds on top of #247. For a cleaner diff see commit https://github.com/jupyterlab/debugger/commit/5df41546222067a75428d8b474384aa789183ca6.

[Binder](https://mybinder.org/v2/gh/jtpio/debugger/local-module-files?urlpath=/lab/tree/examples/index.ipynb)

This lets users set breakpoints in local modules and files. The breakpoints show up with the absolute path to the file in the `Breakpoints` split.

### TODO

- [x] Support optional `path` when setting breakpoints
- [x] Rebase

local-module-breakpoints

### DIFF
--- a/src/handlers/tracker.ts
+++ b/src/handlers/tracker.ts
@@ -138,7 +138,8 @@ export class TrackerHandler implements IDisposable {
     const editor = editorWrapper.editor;
     const editorHandler = new EditorHandler({
       debuggerService: this.debuggerService,
-      editor
+      editor,
+      path
     });
     const widget = new MainAreaWidget<CodeEditorWrapper>({
       content: editorWrapper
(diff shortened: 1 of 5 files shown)


src/handlers/tracker.ts
```diff
@@ -264,7 +265,7 @@ export class TrackerHandler implements IDisposable {
 
       const code = editor.model.value.text;
       const codeId = this.debuggerService.getCodeId(code);
-      if (source !== codeId) {
+      if (widget.title.caption !== source && source !== codeId) {
```
jtpio: This assumes the `caption` is always equal to the full path.

We could think of a better way to do this by maybe adding extra data to the `editor`.

afshin: Or we could use an `AttachedProperty`, potentially. But for now, this seems fine.

PR #260, opened 2019-12-09
Show source path above the read-only preview editor

### LINKED ISSUE #253: [UX] Show Source Path
Given that the source may not point to the active file, I suggest adding a file path to the source heading. See attached image
Debugger source

jtpio:
Fixes #253.

Show the source path as we step into the code, so it's easier to know where the current execution has stopped.

show-source-path

### DIFF
--- a/src/callstack/index.ts
+++ b/src/callstack/index.ts
@@ -74,7 +74,7 @@ class CallstackHeader extends Widget {
     super({ node: document.createElement('header') });
 
     const layout = new PanelLayout();
-    const span = new Widget({ node: document.createElement('span') });
+    const span = new Widget({ node: document.createElement('h2') });
 
     this.layout = layout;
     span.node.textContent = title;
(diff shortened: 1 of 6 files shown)


jtpio:
This does not remove the `uppercase` from the titles in the headers, as this could be addressed separately.

src/callstack/index.ts
```diff
@@ -74,7 +74,7 @@ class CallstackHeader extends Widget {
     super({ node: document.createElement('header') });
 
     const layout = new PanelLayout();
-    const span = new Widget({ node: document.createElement('span') });
+    const span = new Widget({ node: document.createElement('h2') });
```
afshin: Maybe this should be called `title` like the other one.

src/variables/index.ts
```diff
@@ -43,7 +43,7 @@ class VariablesHeader extends Widget {
   constructor(title: string) {
     super({ node: document.createElement('header') });
     const layout = new PanelLayout();
-    const span = new Widget({ node: document.createElement('span') });
+    const span = new Widget({ node: document.createElement('h2') });
```
afshin: Also here...

src/callstack/index.ts
jtpio: Yes that's better.

Issue #171, opened 2019-11-12
Breakpoints UI


jtpio:
The breakpoints view now shows the breakpoints with the name of their corresponding temporary files:

![image](https://user-images.githubusercontent.com/591645/68696891-1b48de80-057e-11ea-96e5-f8aba0c82d39.png)

Some points to consider:

- Should the breakpoints be grouped per cell, and show something from the cell UI such as the execution count or some text (the first line of the cell for example), if possible?
- Should the breakpoints for previously dumped cells be shown in the UI, even if these cells are not reachable anymore? It might not be obvious to do so, since the files exist on disk and corresponding breakpoints are set in `ptvsd`. Sometimes we can also map back to the same files (editing followed by successive `Ctrl-Z`).

martinRenou:
What would be nice would be to add the line number and the corresponding line, like the JavaScript debugger does in the Chrome developer tools UI
![breakpoints](https://user-images.githubusercontent.com/21197331/71000961-884f1700-20dc-11ea-8835-0d1fe33d8ec2.png)

jtpio:
Yes, there was a similar thing in the early prototype:

![](https://user-images.githubusercontent.com/591645/61885305-e563e980-aefd-11e9-9a33-15eeacb73f09.gif)

jtpio:
So this could be added as second line.

JohanMabille:
Yes that would be great, since the names of the temporary files are not really meaningful

PR #262, opened 2019-12-09
Align breakpoint line numbers

jtpio:
A quick pass on the breakpoints UI to:

- align the line numbers to the right
- handle the overflow for long source path
- add title attribute with the full path

image

### DIFF
--- a/src/breakpoints/body.tsx
+++ b/src/breakpoints/body.tsx
@@ -94,6 +94,7 @@ const BreakpointComponent = ({
     <div
       className={`breakpoint`}
       onClick={() => model.clicked.emit(breakpoint)}
+      title={breakpoint.source.path}
     >
       <span className={'marker'}>●</span>
       <span className={'source'}>{breakpoint.source.path}</span>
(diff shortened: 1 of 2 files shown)


afshin:
This is strictly an improvement, so let's merge it.

I think a further enhancement could be applied to the truncating of the file name to truncate at the beginning instead of at the end, like the JupyterLab file browser:

Screen Shot 2019-12-09 at 12 22 16 PM

src/breakpoints/body.tsx
```diff
@@ -94,10 +94,11 @@ const BreakpointComponent = ({
     <div
       className={`breakpoint`}
       onClick={() => model.clicked.emit(breakpoint)}
+      title={breakpoint.source.path}
     >
-      <span>
-        {breakpoint.source.path} : {breakpoint.line}
-      </span>
+      <span className={'marker'}>●</span>
```
afshin: Are the class names `marker` and `source` intentionally not prefixed?

jtpio: For now yes, as it follows the existing style (we should also change the `breakpoint` class from the wrapper `<div>`).

We will have to do a proper pass anyway with #139, but happy to make the here too.

jtpio:
That would probably be better indeed.

We'll have to see if it makes sense to do that now, since there were a couple of ideas and questions for an alternative UI breakpoint (https://github.com/jupyterlab/debugger/issues/171 and others).

PR #266, opened 2019-12-10
Add line highlighting to code consoles

### LINKED ISSUE #265: Debugging in consoles
This is what debugging in consoles looks like at the moment:

debug-console

When executing code with `Shift-Enter`:

- The gutters are sent to the console with the breakpoints
- There is no line highlighting in the console panel
- Breakpoints can be added and removed after the code is sent to the console panel

Maybe a better behavior would be to debug the code *before* it gets sent to the console. But there is probably no reasonable way to block this behavior.

Otherwise it looks like it would also be fine to show the line highlighting in the editors that have been sent to the console panel.

jtpio:
Fixes #265.

Since the consoles keep track of the previous executed cells, we can use a similar approach as for the notebook.

debug-console-highlight


jtpio:
Similar flow as for the notebook when dealing with previous executions:

debug-console-highlight2

jtpio:
This also supports the "Show Kernel Activity" when the kernel is shared between a notebook and a console:

debug-console-activity

PR #271, opened 2019-12-11
Add installation instructions to the README

jtpio:
Now that the alpha has been published we can add some instructions about the pre-release and how to install it.


jtpio:
cc @afshin

PR #273, opened 2019-12-11
Move ensureSessionReady to a ready promise

jtpio:
This is the first part of the wider kernel status handling from #76.

For now we have been using an internal `ensureSessionReady` method to make sure a session (and kernel) is ready before sending messages.

This moves the logic to a `ready` promise which doesn't assume the session is a `IClientSession`.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -1,8 +1,6 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { IClientSession } from '@jupyterlab/apputils';
-
 import { IDisposable } from '@phosphor/disposable';
 
 import { ISignal, Signal } from '@phosphor/signaling';
@@ -307,8 +305,6 @@ export class DebugService implements IDebugger, IDisposable {
     if (!this.session.isStarted) {
       return;
     }
-    // Workaround: this should not be called before the session has started
-    await this.ensureSessionReady();
     if (!path) {
       const dumpedCell = await this.dumpCell(code);
       path = dumpedCell.sourcePath;
@@ -436,8 +432,6 @@ export class DebugService implements IDebugger, IDisposable {
     breakpoints: DebugProtocol.SourceBreakpoint[],
     path: string
   ) => {
-    // Workaround: this should not be called before the session has started
-    await this.ensureSessionReady();
     return await this.session.sendRequest('setBreakpoints', {
       breakpoints: breakpoints,
       source: { path },
@@ -461,11 +455,6 @@ export class DebugService implements IDebugger, IDisposable {
       };
     });
   };
-
-  private async ensureSessionReady(): Promise<void> {
-    const client = this.session.client as IClientSession;
-    return client.ready;
-  }
 
   private clearModel() {
     this._model.callstack.frames = [];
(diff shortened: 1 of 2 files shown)


afshin:
Maybe the ready state of the session should be a private attribute of the session itself. Since its API is all promises, it could automatically `await this._ready;` without any of its clients needing to know.

It actually seems like you're already doing this in the `restoreState()` method ... so waiting for ready right before a client calls that function is superfluous.

jtpio:
The main reason to make it public was to be able to then request the kernel info with `debug.isDebuggingEnabled` which requires the kernel to be ready.

We use the kernel info to know if the kernel has support for debugging, so it could have been useful to have that as part of the API surface to avoid restoring a state for a kernel that doesn't support debugging.

jtpio:
Digging into this a bit more, it turns out that even when `debug.session.ready` resolves, the kernel info is still `null` / not available.

afshin:
Is the kernelspec a place where we can store metadata that says whether the kernel supports debugging? That way we don't need to make a request to know if debugging is available.

jtpio:
There should be a way to store metadata yes, but we probably want to keep it in the `kernel_info_reply` too.

I'll check whether we it's possible to wait for the kernel info reply to arrive. For reference this is where it is requested: https://github.com/jupyterlab/jupyterlab/blob/ea87dd7166c67b05a7f87df3f415e9a4eafdf029/packages/services/src/kernel/default.ts#L1199

jtpio:
> it turns out that even when `debug.session.ready` resolves, the kernel info is still `null` / not available.

Correction: for a `ClientSession`, we must wait for `client.ready` to resolve before accessing `client.kernel` (which is `null` otherwise).

afshin:
Nice! I like this pattern because it gives the client fewer things to worry about and handles readiness internally.

jtpio:
OK so this has now been changed to using an internal `_ready` promise.

There isn't many methods exposed publicly so it didn't require adding too many `await this._ready`.
This should also make it easier to handle kernel status changes (the `_ready` promise could be modified directly).

jtpio:
We'll probably want to change the way we check for debugging capabilities.

For now this has been accessed with a getter, but we'll have to actively `await` it (an internally await `this._ready`) before accessing the kernel info.

This might become something like this instead:

```typescript
const enabled = await debug.requestDebuggingEnabled()
```

Open for other suggestions of course. Also this could be addressed in a different change.

jtpio:
> Nice! I like this pattern because it gives the client fewer things to worry about and handles readiness internally.

Yes it makes things more transparent. Thanks for the suggestion :+1:

afshin:
Hm, `enabled` would always be a promise in that case, though?

jtpio:
Or rather the value of the promise?

Internally the service will have to do something like this:

```typescript
async isDebuggingEnabled(): Promise<boolean> {
  const kernelInfo = await this._session.requestKernelInfo(); 
  return kernelInfo?.debugger ?? false;
}
```

And `this._session.requestKernelInfo()` would also return a promise (doing an `await this._ready` under the hood).

jtpio:
#275 builds on top of this and implements this idea (commit https://github.com/jupyterlab/debugger/pull/275/commits/122d2d11ec2c417306ed1299a33836c81ce22fe9)

PR #275, opened 2019-12-11
Check debugging is enabled before restoring the session

### LINKED ISSUE #268: Canceled promise error for kernels without support for debugging
For now we still create a new debug session even if the kernel doesn't support debugging.

After the debug session is created, we send a `debugInfo` message and wait for the reply on the `control` channel. For non-debugging kernels, this reply never comes.

ipykernel-debuginfo

Logging this as an issue so we can keep track of it and make sure the debugger doesn't interfere with regular workflows that use `ipykernel`.

This should be fixed by #239.

jtpio:
Fixes #268.

Check whether a kernel support debugging before restoring the session and sending the `debugInfo` message.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -66,7 +66,7 @@ export class DebugSession implements IDebugger.ISession {
       this._client.iopubMessage.connect(this._handleEvent, this);
 
       if (this.client instanceof ClientSession) {
-        this._ready = this.client.ready;
+        this._ready = this.client.ready.then(_ => this.client.kernel.ready);
       } else {
         this._ready = this.client.kernel.ready;
       }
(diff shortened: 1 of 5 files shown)


jtpio:
This does not support changing kernels for existing notebook, which should be handled in #76.

src/session.ts
```diff
@@ -62,20 +62,24 @@ export class DebugSession implements IDebugger.ISession {
 
     this._client = client;
 
-    if (client) {
+    if (this._client) {
       this._client.iopubMessage.connect(this._handleEvent, this);
+
+      if (this.client instanceof ClientSession) {
+        this._ready = this.client.ready.then(_ => this.client.kernel.ready);
```
jtpio: This is required to be able to retrieve the kernel info.

jtpio:
This approach will probably slightly change when we'll have to also handle kernel changes:

image

jtpio:
cc @afshin 

Let me know what you think. Then we can see how we can make everything fit together with #239 (or after it lands).

As mentioned before, the main challenge here is the need to wait for the kernel to be ready before accessing the kernel info (and check for debugging support).

PR #276, opened 2019-12-12
Handle kernel restarts

### LINKED ISSUE #76: Handle kernel status

jtpio:
Fixes #76.

Restarting the kernel will also restart the debugger. For now the breakpoints are not saved and restored.

This needs to take into account the state of the toggle button from #239 later on. 

kernel-restart

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -68,40 +68,21 @@ export namespace CommandIDs {
   export const debugConsole = 'debugger:debug-console';
 }
 
-async function setDebugSession(
-  app: JupyterFrontEnd,
-  debug: IDebugger,
-  client: IClientSession | Session.ISession
-) {
-  if (!debug.session) {
-    debug.session = new DebugSession({ client: client });
-  } else {
-    debug.session.client = client;
-  }
-  const debuggingEnabled = await debug.requestDebuggingEnabled();
-  if (!debuggingEnabled) {
-    return;
-  }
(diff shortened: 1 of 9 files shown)


src/index.ts
```diff
@@ -71,9 +71,12 @@ async function setDebugSession(
   client: IClientSession | Session.ISession
 ) {
   if (!debug.session) {
-    debug.session = new DebugSession({ client: client });
-  } else {
-    debug.session.client = client;
+    debug.session = new DebugSession();
+  }
+  debug.session.client = client;
```
jtpio: This follows the same approach as when we set the session on the service.

It removes the possibility to provide a `client` to the constructor, so the client can only be set via the setter.

src/service.ts
```diff
@@ -78,6 +83,30 @@ export class DebugService implements IDebugger, IDisposable {
...
       this._eventMessage.emit(event);
     });
+
+    this._session?.clientChanged.connect(
+      (_: IDebugger.ISession, client: IClientSession | Session.ISession) => {
+        if (!client) {
+          return;
+        }
+        // TODO: listen to status changes only if the client / kernel
+        // has support for debugging
```
jtpio: This could be done at the plugin level, re-using some of the logic from #239

PR #278, opened 2019-12-12
Dispose the session on debugger closed

jtpio:
The session can be disposed before setting it to `null` when the debugger is closed.


jtpio:
This is ready to go.

PR #279, opened 2019-12-12
Fix enabled commands on debugger closed

jtpio:
Commands were not properly disabled when the debugger sidebar was closed.

This change should fix it.

### Before

close-debugger-update-commands2

### After

close-debugger-update-commands

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -147,7 +147,7 @@ export class DebugService implements IDebugger, IDisposable {
    * Whether there exists a thread in stopped state.
    */
   hasStoppedThreads(): boolean {
-    return this._model?.stoppedThreads.size !== 0 ?? false;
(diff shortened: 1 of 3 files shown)


src/service.ts
```diff
@@ -147,7 +147,7 @@ export class DebugService implements IDebugger, IDisposable {
    * Whether there exists a thread in stopped state.
    */
   hasStoppedThreads(): boolean {
-    return this._model?.stoppedThreads.size !== 0 ?? false;
+    return this._model?.stoppedThreads.size > 0 ?? false;
```
jtpio: This was an issue because when `this._model` is `null`, we would perform `null !== 0`, which is `true`.

jtpio:
We might make some changes to the lifecycle of the sidebar later on, but for now this should already be an improvement.

PR #281, opened 2019-12-13
Update service.ts to keep a consistent style

jtpio:
A quick pass on `service.ts` to:

- [x] keep a consistent way to declare methods
- [x] make methods `private` if not accessed elsewhere
- [x] rename some methods to better match with what they actually do
- [x] add docstring to `public` methods
- [x] add docstring to `private` methods
- [x] follow #139


jtpio:
This `service.ts` file should now follow #139 pretty closely.

PR #290, opened 2019-12-16
Handle debugger lifecycle

### LINKED ISSUE #289: Toggle button and sidebar visibility
*Originally posted in https://github.com/jupyterlab/debugger/pull/239#issuecomment-565490444*.

If we want to be able to manage the lifecycle of the debugger and the sidebar separately, we need to handle a few other cases:

- Automatically create the sidebar when switching the button on
- Closing the sidebar should switch the buttons off.

At first it sounds reasonable to always show the toggle buttons if the kernel supports debugging.

At the moment we can end up in situations like the following:

debugger-sidebar-lifecycle

### LINKED ISSUE #292: Debugger automatically starts then stops when new notebook is opened.
It looks as though the debugger is automatically started and then stopped without the user intervening.

Screen Shot 2019-12-16 at 1 53 20 PM

jtpio:
Fixes #292.

A couple of fixes as a follow-up to #239. 

To try it out:

[Binder](https://mybinder.org/v2/gh/jtpio/debugger/button-sidebar?urlpath=/lab/tree/examples/index.ipynb)

### TODO

- [x] Turn debugging off when the sidebar is closed
- [x] Notebook and console handlers lifecycle should follow the toggle button
- [x] ~Create the sidebar when enabling debugging via the button if it does not exist yet. Otherwise: do not show the switch button if there is no sidebar?~ (Can be handled separately)

### Keep last debug session

toggle-last-session

### Handle widgets with the same session

toggle-sync-sessions

### Restore the state from the kernel

Remove the recently added `debuggedClients` to stay stateless

toggle-restore-state-from-kernel


jtpio:
To make the experience less confusing and avoid starting and stopping the debugger, we should tie the handlers lifecycle to the toggle button, and whether debugging is enabled for a given session path:

toggle-dispose-handlers

jtpio:
It turns out that the toggle button logic needed quite some refactoring. The main takeaway is that toggling the button should also create and dispose the debug handlers properly.

This change also removes the `debuggedClient` state that was introduced in #239, so the debugger still stays stateless.

To try it out:

[Binder](https://mybinder.org/v2/gh/jtpio/debugger/button-sidebar?urlpath=/lab/tree/examples/index.ipynb)

As an improvement, we could consider adding a signal to the debug session, so the UI can update itself when the state of the session changes (started or stopped). For now we have to focus back on the widget to have the UI changes propagated.

jtpio:
The second main change is that we now don't create a `DebugSession` anymore for a `client` that doesn't support debugging.

This allows the debugger to keep "focus" on the previous `client` that supports debugging. As a side-effect we can inspect variables for a notebook being debugged, while working on a notebook that doesn't have support for debugging:

toggle-last-session

src/index.ts
```diff
@@ -114,86 +95,123 @@ async function updateToolbar(
...
   constructor(builder: new (option: any) => H) {
-    this.builder = builder;
+    this._builder = builder;
+  }
+
+  /**
+   * Dispose all the handlers.
+   * @param debug The debug service.
+   */
+  async disposeAll(debug: IDebugger) {
```
afshin: Does this need to be `async`? Also, is this an appropriate place to use the `DisposableSet` from `@lumino/disposable`?

src/session.ts
```diff
@@ -238,4 +224,19 @@ export namespace DebugSession {
...
     client: IClientSession | Session.ISession;
   }
+
+  /**
+   * Request whether debugging is enabled for the given client.
+   * @param client The client session.
+   */
+  export async function requestDebuggingEnabled(
+    client: IClientSession | Session.ISession
+  ) {
```
afshin: This function should have a return type: `Promise<boolean>` and a return statement that guarantees a boolean is returned: `return !!(info?.debugger ?? false);`

src/index.ts
jtpio: Good question.

This method might not even be needed if we decide to drop the `close` command: https://github.com/jupyterlab/debugger/issues/295

afshin: If you don't have any kernels that support debugging, you should probably never see a debugger icon/sidebar. It does seem like if you want your sidebar back as empty, you might be annoyed that you can't get rid of the debugger. I'm not sure which way to go about this. Since the right sidebar is empty by default, I'm ambivalent. What do you think?

src/session.ts
jtpio: Sounds good :+1: 

It's true that kernels can still return an empty string for example for the `debugger` key.

afshin: Sure, but we will have to assume that's `false`.

This might belong in the `IDebugger` interface so others can consume it, as a static method in that namespace ... `IDebugger#isAvailable(client)` maybe? It seems like the kind of thing consumers of the `IDebugger` might care about?

jtpio: > It's true that kernels can still return an empty string for example for the `debugger` key.

I realize that sentence was maybe a bit confusing. What I meant is that we indeed need to make sure the return type is a `boolean` because for now the `debugger` key is not part of the protocol yet, so we could get other kinds of values. So +1 to enforce the `boolean` type.

> This might belong in the `IDebugger` interface so others can consume it, as a static method in that namespace ... `IDebugger#isAvailable(client)` maybe? It seems like the kind of thing consumers of the `IDebugger` might care about?

That could indeed be part of the public API.

src/index.ts
jtpio: It's probably something we want to discuss in https://github.com/jupyterlab/debugger/issues/295.

But it does feel like having 2 "on / off" states puts some cognitive load on the user. Maybe the extension can be disabled via the extension manager to remove the sidebar?

jtpio:
Will do a quick update to:

- [ ] move `DebugSession#requestDebuggingEnabled` to `IDebugger#isAvailable`
- [ ] guarantee `IDebugger#isAvailable` return a `boolean`

src/index.ts
afshin: Perhaps there should be a show/hide toggle in the `View` menu that is *available* only when the debugger is in the sidebar and is *enabled* only if the debugger is not turned on.

You're right, it's multiple layers, but I think if it's in the view menu, it will both be discoverable and a little less confusing. It won't be an explicit thing in a user's field of vision.

KsavinN:
Don't not why but I see that button changed shape to a circle.
dotNotToggle

src/index.ts
jtpio: > is _enabled_ only if the debugger is not turned on.

Does "turned on" in that case mean "the toggle button is on"?

src/session.ts
jtpio: It looks like TS would not allow defining a static method (exported function) in that namespace if we try to follow that pattern: 

https://www.typescriptlang.org/play/index.html?ssl=15&ssc=6&pln=13&pc=1#code/CYUwxgNghgTiAEkoGdnwCoHsDWIB2APOgHzwDeAsAFDy2KZ7IAuMArmE5jABR5QC2IAFzxmMAJZ4A5gEoA3NQC+1FVRAAPAA5cm8SUxAwAZlDAIAkgBEQAI1ZSph8tTrw4UYAwgBPPcgDKTLAGwCI2mJgQIFB4CjR0zMHcMiIACjCY-OLIIAQAbpjiwMRxrsyYmslpGVk5+YXFccpU1BraMLp8gsiaphbWdg5OlPG0bTrwRqx4HOIMfgCCeVDi0DZRyc6jrm4gTKwwePAsrCCldM3NrVoTYAzM8Fa29o4w8AC88HggAO4YOPgCE9Bq9iNwAEQAAQAVqxNN4DDA1gB6UAgwzg+RAA

afshin:
Hm, @KsavinN, try running a `git clean -fdx` before rebuilding, maybe? I see the correct toggle when I build it.

KsavinN:
> Hm, @KsavinN, try running a `git clean -fdx` before rebuilding, maybe? I see the correct toggle when I build it.

I did it locally. It's work properly. That bug appears during using a binder.

jtpio:
Also seeing that on Binder, but not locally.

afshin:
@jtpio I guess it should just be a method on `DebugService` that is exposed on the `IDebugger` interface.

jtpio:
That sounds like the best option indeed.

jtpio:
For Binder and the toggle button, it looks like an issue with JupyterLab 1.1.4 (which is used on Binder for now).

This can be reproduced locally in a fresh environment:

```
conda create -n debugger-test -c conda-forge jupyterlab=1.1.4 xeus-python ptvsd
jupyter labextension install @jupyterlab/debugger --debug
jupyter lab xpython_example.ipynb
```

We are probably all using 1.2+ locally.

src/session.ts
jtpio: Done as a method in `DebugService` and exposed via `IDebugger`.

src/index.ts
jtpio: To come back to the first comment, `disposeAll` doesn't need to be `async` indeed (removed in the latest commit).

jtpio:
This should be ready to go now, cc @afshin @KsavinN if you want to have another look.

PR #291, opened 2019-12-16
Breakpoints removed from the sidebar on notebook closed

KsavinN:
Resolved by Jeremy in PR #308 

**This branch needs to be rebased and refactored after this PR #290** 

resolve issue #168.

- [x] after close all notebooks clear breakpoints.

After close widget:

clearBreakpoints

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -300,6 +300,7 @@ const notebooks: JupyterFrontEndPlugin<void> = {
 
     labShell.currentChanged.connect(async (_, update) => {
       const widget = update.newValue;
+
       if (!(widget instanceof NotebookPanel)) {
         return;
       }
--- a/src/index.ts
+++ b/src/index.ts
@@ -179,6 +179,11 @@ class DebuggerHandler<
       handler.dispose();
       delete this._handlers[widget.id];
       updateAttribute();
+      const handlerIds = Object.keys(this._handlers);
+      if (handlerIds.length !== 0) {
+        return;
+      }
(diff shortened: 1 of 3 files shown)


jtpio:
Should we really clear the breakpoints from the state in the kernel? Or just from the UI?

If we re-open the notebook the breakpoints will be gone, even if the kernel was not shut down.

KsavinN:
> Should we really clear the breakpoints from the state in the kernel? Or just from the UI?
> 
> If we re-open the notebook the breakpoints will be gone, even if the kernel was not shut down.

It should still have breakpoints.  It should only clear UI.
stillSHow

src/index.ts
```diff
@@ -125,6 +125,8 @@ async function updateToolbar(
           debug.session.debuggedClients.add(debug.session.client.path);
           await debug.start();
         } else {
+          await debug.restart();
+          await debug.clearBreakpoints();
```
jtpio: Not sure we should clean the breakpoints separately, as this looks like a special case.

jtpio:
This shouldn't be specific to breakpoints only. There is a similar issue when the execution is stopped and the notebook is closed:

dispose-on-execution-stopped

Maybe there could be a more generic model cleanup method that would be triggered when the widget is disposed? The sidebar would react to model changes and clean itself up.

jtpio:
Ideally this would also handle kernel changes, so the sidebar UI would be cleared when the kernel is changed:

clear-sidebar-kernel-changed

PR #298, opened 2019-12-17
Remove close command

### LINKED ISSUE #295: Remove the close command
Should the `close` command be removed, so the sidebar is always created and never closed?

While looking at #290, it started to feel a bit confusing to close the debugger sidebar while debugging. That also means users have to keep a mental model that consist of different states:

- the debugger sidebar is open / closed
- the toggle button is on / off 

Removing the possibility to close the sidebar would simplify things greatly. There is only one way to turn on debugging and that is by switching the button on in a notebook, console or file.

For the users who have installed the extension but don't want to use it anymore, they could disable it from the extension manager and that would remove the sidebar.

KsavinN:
resolve #295

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -66,8 +66,6 @@ export namespace CommandIDs {
   export const stepOut = 'debugger:stepOut';
 
   export const debugConsole = 'debugger:debug-console';
-
-  export const close = 'debugger:close';
 }
 
 async function setDebugSession(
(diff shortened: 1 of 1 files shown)


jtpio:
Should we also remove the create from the palette?

jtpio:
Ah too late.

jtpio:
Because the sidebar is created automatically during the plugin initialization.

KsavinN:
we can remove that in separate commit will look better in the commit history.

PR #301, opened 2019-12-17
Remove create command 

### LINKED ISSUE #300: Remove the create command from the palette
The `create` command doesn't need to be shown in the palette for now, since the ability to close the sidebar was removed in #298.

image

KsavinN:
resolve #300 

- [x] remove create command from the palette.

- [x] remove command create and create sidebar directly.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -422,7 +422,6 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
     });
 
     commands.addCommand(CommandIDs.create, {
-      label: 'Debugger',
       execute: async () => {
         if (sidebar) {
           return;
@@ -466,7 +465,6 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
     if (palette) {
       const category = 'Debugger';
       [
-        CommandIDs.create,
         CommandIDs.debugContinue,
         CommandIDs.terminate,
         CommandIDs.next,
--- a/src/index.ts
+++ b/src/index.ts
@@ -49,8 +49,6 @@ import { IDebugger } from './tokens';
  * The command IDs used by the debugger plugin.
  */
 export namespace CommandIDs {
-  export const create = 'debugger:create';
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -421,52 +419,43 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
...
-        });
-
-        sidebar.service.sessionChanged.connect(_ => {
-          commands.notifyCommandChanged();
-        });
-
-        if (labShell.currentWidget) {
-          labShell.currentWidget.activate();
-        }
+    // Createing sidebar instance and add to shell
```
jtpio: It's probably also fine without a comment here.

src/index.ts
```diff
@@ -421,52 +419,43 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
...
 
-        shell.add(sidebar, 'right');
-      }
+    sidebar.service.sessionChanged.connect(_ => {
+      commands.notifyCommandChanged();
     });
 
+    if (labShell.currentWidget) {
+      labShell.currentWidget.activate();
+    }
```
jtpio: Wondering whether we still need this actually, or if it can be removed?

src/index.ts
```diff
@@ -421,52 +419,38 @@ const main: JupyterFrontEndPlugin<IDebugger> = {
...
-        });
-
-        sidebar.service.sessionChanged.connect(_ => {
-          commands.notifyCommandChanged();
-        });
-
-        if (labShell.currentWidget) {
-          labShell.currentWidget.activate();
-        }
+    sidebar = new Debugger.Sidebar({
```
jtpio: One last thing: we could declare and define the `sidebar` variable here directly, with:

```typescript
const sidebar = new Debugger.Sidebar({...})
```

KsavinN: rigth

PR #302, opened 2019-12-17
Handle kernel changes

### LINKED ISSUE #277: Handle kernel changes
We should handle kernel changes for a given session. For example when a user decides to switch from `ipykernel` to `xeus-python` (or the other way around) with the kernel selector popup:

image

This is not exactly the same as #76. #76 is about kernel status changes.

jtpio:
Fixes #277.

To try it out:

[Binder](https://mybinder.org/v2/gh/jtpio/debugger/kernel-changed?urlpath=/lab/tree/examples/index.ipynb)

### TODO

- [x] Handle notebooks
- [x] Handle consoles
- [x] Handle files?

kernel-changes

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -95,6 +95,7 @@ function updateToolbar(
   if (itemAdded && widget instanceof ConsolePanel) {
     widget.insertWidget(0, toolbar);
   }
+  return button;
 }
 
 /**
@@ -129,7 +130,8 @@ class DebuggerHandler<
   }
 
   /**
-   * Update a debug handler for the given widget.
+   * Update a debug handler for the given widget, and
+   * handle kernel changed events.
    * @param debug The debug service.
    * @param widget The widget to update.
    */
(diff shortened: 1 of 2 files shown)


jtpio:
This should also handle the case when a console is created from a notebook:

kernel-changes-consoles

In the screencast above we see that the line numbers are shown in the console panel when the kernel is `ipykernel`. This is because at the moment the handler is disposed only when the widget gets the focus. This goes back to some observations made in another PR about keeping the handlers in sync or listening to state changes.

jtpio:
As well as shutting down the kernels (since it goes through the "No Kernel" state):

kernel-changes-shutdown

jtpio:
cc @afshin @KsavinN if you want to give it a try

afshin:
This works for me but appears to not get rid of breakpoints when I switch.

switch kernel

jtpio:
Indeed.

It's still related to the way the sidebar UI should be cleared (similar to when the notebook is disposed, cf #168),

afshin:
Okay, so outside of the scope of this PR, then?

Otherwise, this looks good to me.

jtpio:
Yes let's see if we can handle that in #291.

PR #304, opened 2019-12-18
Handle kernel restarts

### LINKED ISSUE #76: Handle kernel status

jtpio:
Fixes #76, replaces #276.

kernel-restart2

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -140,15 +140,40 @@ class DebuggerHandler<
     widget: W,
     client: IClientSession | Session.ISession
   ): Promise<void> {
+    const updateHandler = async () => {
+      return this._update(debug, widget, client);
+    };
+
+    // setup handler when the kernel changes
     const kernelChangedHandler = this._kernelChangedHandlers[client.path];
     if (kernelChangedHandler) {
       client.kernelChanged.disconnect(kernelChangedHandler);
     }
-    const updateHandler = async () => {
-      return this._update(debug, widget, client);
-    };
-    client.kernelChanged.connect(updateHandler, this);
+    client.kernelChanged.connect(updateHandler);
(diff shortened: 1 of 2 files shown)


jtpio:
For now the `restarting` status seems to be emitted *before* the kernel is actually restarted.

This means that the handler triggers the restore flow before the restart, and the `debugInfo` returns `isStarted: true`.

jtpio:
This should also handle kernels being killed and automatically restarted:

kill-kernel-restarts

src/index.ts
```diff
@@ -138,20 +140,61 @@ class DebuggerHandler<
...
+      // wait for the first `idle` status after a restart
+      if (restarted && client.status === 'idle') {
+        restarted = false;
+        return updateHandler();
+      }
+      // handle `starting`, `restarting` and `autorestarting`
+      if (client.status.endsWith('starting')) {
+        restarted = true;
+      }
+    };
```
jtpio: Not super satisfied with this logic.

It seems to be a bit tricky to really detect a kernel restart.

jtpio: An alternative would be to schedule the call to `updateHandler` with a small time delay.

jtpio:
Rebased.

jtpio:
cc @afshin @KsavinN if you want to try it.

This should handle most of the restart scenarios, for example a click on the restart button or a kernel killed from the command line.

However the logic to properly detect a kernel restart is a bit convoluted, and it would be great to have a second look at it later on if we decide to keep it like this for now.

jtpio:
Yes, I think we'll need to revisit the logic a bit.

Thanks for looking at it :+1:

PR #308, opened 2019-12-19
Clear the sidebar UI

### LINKED ISSUE #168: Breakpoints not removed from the sidebar on notebook closed

jtpio:
Fixes #168.

Alternative to #291. 

This introduces a new `clear()` method on the model to clear everything, and restore the breakpoints to an empty map.

This should also still support keeping the last debug session alive when switching between notebooks and widgets.

clear-sidebar-ui

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -144,12 +144,12 @@ class DebuggerHandler<
     if (kernelChangedHandler) {
       client.kernelChanged.disconnect(kernelChangedHandler);
     }
-    const updateHandler = () => {
-      void this._update(debug, widget, client);
+    const updateHandler = async () => {
+      return this._update(debug, widget, client);
     };
     client.kernelChanged.connect(updateHandler, this);
     this._kernelChangedHandlers[client.path] = updateHandler;
-    return this._update(debug, widget, client);
+    return updateHandler();
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -194,6 +194,14 @@ class DebuggerHandler<
       }
       handler.dispose();
       delete this._handlers[widget.id];
+
+      // clear the model if the handler being removed corresponds
+      // to the current active debug session
+      if (debug.session?.client?.path === client.path) {
+        const model = debug.model as Debugger.Model;
+        model.clear();
```
jtpio: There could be a call to `debug.stop()` instead, so this would also handle the case when the execution has stopped.

But when the notebook is disposed, its kernel (via `client.kernel`) is already set to `null`, which prevents sending the `disconnect` message.

Also, stopping the debugger wouldn't play nice when two notebooks are using the same kernel.

jtpio:
cc @KsavinN let me know what you think.

This takes a slightly different approach than #291 and tries to not be too specific to breakpoints.

KsavinN:
It looks definitely better. I will test that locally :)

jtpio:
Cool :)

The tricky part about clearing the model is that we still want to keep track of the previous active debug session (to enable debugging a notebook while looking at another one), if available.

KsavinN:
Tested Locally, we can merge it

PR #310, opened 2019-12-19
Format the handlers according to the code style

jtpio:
Part of #139.

Similar to previous PRs, this one is about formatting the handlers so the style stays consistent.


jtpio:
This could go in as is right now (it's mostly about adding docstrings and renaming) but I'm also fine to hold it until after 0.1.0.

PR #323, opened 2019-12-23
Check for widget focus when toggling debugging

### LINKED ISSUE #318: Focus issues with the toggle button
When there is a split with 2 notebooks, we a user clicks on the toggle button for the notebook that is *doesn't* have the focus, the debug session for the notebook with the focus is the one that is stopped.

Or put it simply:

toggle-button-focus-issue

The handler for the switch should make sure the widget has focus before switching.

jtpio:
Fixes #318. 

toggle-button-focus-fixed

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -136,12 +136,13 @@ class DebuggerHandler<
    * @param widget The widget to update.
    */
   async update<W extends ConsolePanel | NotebookPanel | DocumentWidget>(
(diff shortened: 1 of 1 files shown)


jtpio:
In the screencast above, the first click brings the focus back to the notebook. The second click goes through as the focus (`currentWidget`) has changed.

jtpio:
cc @KsavinN if you want to have a look

PR #326, opened 2019-12-24
Move DebuggerHandler to a separate file

### LINKED ISSUE #324: Move DebuggerHandler to its own file
The `DebuggerHandler` has been growing quite a bit recently with the addition of the toggle button logic, as well as kernel restarts and changes.

We could move it to its own file (in the `handlers` folder), and take the opportunity to refactor slightly.

Another thing to consider is to re-add the `SessionTypes` to designate the list of widgets that can be debugged as strings (`notebook`, `console` and `file`), instead of using generics.

jtpio:
Fixes #324.

Progressively transition the `DebuggerHandler` to its own file, and explore changing the generic instantiation approach.

### TODO

- [x] ~Move arrow functions to class methods~ -> To reconsider later


jtpio:
This PR should now address the original intent of moving the `DebuggerHandler` to a separate file.

As discussed in a previous call, this also switches back to using a `SessionType` string when instantiating the handler (`notebook`, `console` and `file`).

jtpio:
cc @afshin since we previously talked about using a `type` to instantiate new handlers.

jtpio:
Note: we could move the `handler.ts` file to the top-level instead.

jtpio:
Opened https://github.com/jupyterlab/debugger/issues/338 as a next improvement to consider for the debugger handlers.

Issue #299, opened 2019-12-17
Show the name of the current debug session in the sidebar


jtpio:
Since it's possible to not have the focus on a notebook, console or file and still have the sidebar show some debug information, we could add an entry in the sidebar to indicate which session is being debugged:

![image](https://user-images.githubusercontent.com/591645/71011875-75454280-20ee-11ea-8528-2b412c0c63a8.png)

Maybe a toolbar at the top of the sidebar?

jtpio:
Fixed by #331.

PR #331, opened 2020-01-03
Create header for sidebar debugger

KsavinN:
related #299 

currentClient

### DIFF
--- a/src/debugger.ts
+++ b/src/debugger.ts
@@ -3,7 +3,7 @@
 
 import { IEditorServices } from '@jupyterlab/codeeditor';
 
-import { SplitPanel } from '@phosphor/widgets';
+import { Panel, SplitPanel } from '@phosphor/widgets';
 
 import { Breakpoints } from './breakpoints';
 
@@ -14,6 +14,8 @@ import { DebuggerModel } from './model';
 import { DebuggerService } from './service';
 
 import { Sources } from './sources';
+
+import { SidebarHeader } from './header';
 
 import { IDebugger } from './tokens';
 
@@ -26,7 +28,7 @@ export namespace Debugger {
   /**
    * A debugger sidebar.
    */
-  export class Sidebar extends SplitPanel {
+  export class Sidebar extends Panel {
     /**
      * Instantiate a new Debugger.Sidebar
      * @param options The instantiation options for a Debugger.Sidebar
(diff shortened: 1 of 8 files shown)


src/debugger.ts
```diff
@@ -114,6 +123,10 @@ export namespace Debugger {
      * The sources widget.
      */
     readonly sources: Sources;
+
+    readonly layoutSplit: SplitPanel;
```
afshin: I would suggest calling this `readonly body` instead of `readonly layoutSplit` since it's more semantically clear.

src/header.ts
```diff
@@ -0,0 +1,29 @@
+import { PanelLayout, Widget } from '@phosphor/widgets';
+
+import { DebuggerService } from './service';
+
+export class SidebarHeader extends Widget {
```
jtpio: Let's try to add docstrings whenever we add new classes / methods?

jtpio: cf #139

src/debugger.ts
jtpio: Also, does it need to be public, or event an attribute to the class?

src/debugger.ts
```diff
@@ -114,6 +123,10 @@ export namespace Debugger {
...
+
+    readonly header: SidebarHeader;
```
jtpio: Do we need this as an attribute?

Or could it be created directly like:

```typescript
const header = new SidebarHeader(...)
this.addWidget(header);
```

src/header.ts
jtpio: The structure of the docstring will be similar to the one used for the other classes.

src/debugger.ts
```diff
@@ -95,6 +106,11 @@ export namespace Debugger {
      */
     readonly service: DebuggerService;
 
+    /**
+     * The body of widget.
+     */
+    readonly body: SplitPanel;
```
afshin: Yeah, actually I think I agree with @jtpio ... this should probably just be `const body = new SplitPanel()` as well ... it doesn't need to be public.

In some future PR would should probably store the relative sizes of the split panel for restoring state on refresh (but not now).

jtpio: > In some future PR would should probably store the relative sizes of the split panel for restoring state on refresh (but not now).

That would definitely be useful :+1: 
Especially if a user wants to collapse some of the sections in the sidebar: https://github.com/jupyterlab/debugger/issues/325

jtpio:
Should the session name also be truncated (same as for the breakpoint paths)?

image

If so we could create a more generic CSS class that would apply to both the breakpoint paths and the session name at the top of the sidebar.

jtpio:
Thanks @KsavinN that looks good.

Do you have a screenshot of what it looks like when the session name is truncated?

KsavinN:
simple

jtpio:
Thanks.

We might want to make the header slightly smaller later but for now this is already an improvement :+1:

Issue #322, opened 2019-12-23
Better way to handle kernel restarts


jtpio:
This is an issue for the following TODO:

https://github.com/jupyterlab/debugger/blob/67441e7f254e7077f4c9983daf73c441170a4c68/src/index.ts#L156

When we update to target JupyterLab 2.0, we might want to have a look at `ConnectionStatus` and the `connecting` status to handle kernel restarts:

https://github.com/jupyterlab/jupyterlab/blob/aa09c52b5e6bd2300dd705a784d4f1eb53b40254/packages/services/src/kernel/kernel.ts#L886

We will also have to update to using a `SessionContext` instead of a `ClientSession`.

jtpio:
Related PR: https://github.com/jupyterlab/jupyterlab/pull/7252

PR #337, opened 2020-01-13
Update to JupyterLab 2.0

### LINKED ISSUE #208: Switch to Lumino
We should eventually switch from `phosphor` to `lumino`: https://github.com/jupyterlab/lumino

Maybe we can wait for the change to happen in JupyterLab first: https://github.com/jupyterlab/jupyterlab/issues/7534

### LINKED ISSUE #336: Upgrade to JupyterLab 2.0
Listing the different steps to target JupyterLab 2.0:

- Update the `@jupyterlab/...` packages to `-beta.2`
- Switch to lumino: #208 
- Services architecture rework (rename `ClientSession` to `SessionContext`): https://github.com/jupyterlab/jupyterlab/pull/7252
- Remove Console specific toolbar logic (`MainAreaWidget`): https://github.com/jupyterlab/debugger/blob/793492b1c3163d8c2a65ee51beebd64b696f0eac/src/index.ts#L83-L89

jtpio:
Fixes #336. 
Fixes #208. 

### Summary of the changes

Here is a list of all the changes, so this PR can also serve as reference when updating other third-party extensions to JupyterLab 2.0

- [x] Update README.md to install JupyterLab 2.0 beta
- [x] Switch to Lumino: #208:
  - This is mostly a search / replace across the code base.
  - It is also useful to remove the existing `yarn.lock` file and `node_modules` folder.
- [x] Rename `iconClassName` to `iconClass`
- [x] Access `toolbar` from the console `MainAreaWidget` to avoid special case for the `ConsolePanel`.
- [x] Fix tests config to handle `svg`
  -  See https://github.com/jupyterlab/jupyterlab/commit/acd909b7ab0950e28933efd3db4ca1afa9c61717#diff-e0a6c733b0412c3c723f4e521c837291 for more info
- [x] Handle breaking changes from the kernel refactoring
  - Go through the changes in https://github.com/jupyterlab/jupyterlab/pull/7252/files
  - One notable change: `await client.kernel.ready` becomes `await sessionContext.session.kernel.info` to check when a kernel is ready
- [x] Update other dependency version numbers to match JupyterLab


jtpio:
Getting there :)

image

jtpio:
This also requires adding the `jest-raw-loader` as the `transform` for `svg` files.

See https://github.com/jupyterlab/jupyterlab/commit/acd909b7ab0950e28933efd3db4ca1afa9c61717#diff-e0a6c733b0412c3c723f4e521c837291 for more info

src/index.ts
```diff
@@ -137,7 +120,7 @@ class DebuggerHandler<
     shell: JupyterFrontEnd.IShell,
     debug: IDebugger,
     widget: W,
-    client: IClientSession | Session.ISession
+    client: Session.ISessionConnection
```
afshin: This is so much nicer ❤️ 

Thanks @jasongrout!

style/icons.css
```diff
@@ -109,8 +109,7 @@
   width: 35px;
 }
 
-.jp-ToggleSwitch::before {
-  content: '';
+.jp-Toolbar-item .jp-ToggleSwitch svg {
```
jtpio: There is now a default `svg` added to the dom when creating a new `ToolbarButton`:

image

We might want to investigate using the new `JLIcon` at some point instead of manipulating the element CSS directly.

afshin: Agreed.

src/session.ts
```diff
@@ -217,10 +211,13 @@ export class DebugSession implements IDebugger.ISession {
...
+  /**
+   * Instantiation options for a `DebugSession`.
+   */
   export interface IOptions {
     /**
-     * The client session used by the debug session.
+     * The session connection used by the debug session.
      */
-    client: IClientSession | Session.ISession;
+    sessionConnection: Session.ISessionConnection;
```
jtpio: Named `sessionConnection` for now to be explicit about what it is.

We could also rename it to just `connection` for succinctness.

afshin: I am always a fan of succinctness.

jtpio: :+1:

jtpio:
Tests pass locally but are failing on CI. Investigating.

jtpio:
Updated the summary of the changes in the [first comment](https://github.com/jupyterlab/debugger/pull/337#issue-362076547), so this can serve as reference for other extensions and also help checking the JupyterLab changelog for 2.0.

jtpio:
There is now a warning logged to the dev tools console:

image

This is probably be a side-effect of the bump to React 16.9 in https://github.com/jupyterlab/jupyterlab/pull/7661, and seems to originate from `react-inspector` (that defines the `TreeView` component).

We might want to give a second try at upgrading `react-inspector` to the latest version (4.0.0) in a separate change.

jtpio:
Marking this one as ready for review.

Two things to investigate after as follow-ups:

- Switching to and from a "No Kernel" state seems to be broken, and requires changing the current active shell widget to trigger the update. We should handle the `kernelChanged` signal at the session context level for the notebook and console panels (instead of session connection level).
- Check whether there is a better way to handler kernel restarts (#322)

We can take a look at these two points separately to keep the current PR focused on migrating to JupyterLab 2.0 only.

cc @afshin

jtpio:
@afshin let's do a `0.1.2` release before merging this one?

jtpio:
Tests seem to be flakier than before. But this might as well be an issue with the tests and the kernel readiness not being awaited correctly.

jtpio:
Adding a missing `await sessionContext.session.kernel.info` like in https://github.com/jupyterlab/jupyterlab/blob/master/tests/test-apputils/src/toolbar.spec.ts#L321 should do the trick.

afshin:
A clean build is failing for me, have you seen this?

```shell
git clean -fdx
pip uninstall jupyterlab
pip install --upgrade --pre jupyterlab
jlpm install
jlpm build
jupyter labextension uninstall @jupyterlab/debugger
jupyter labextension install .
juoyter lab
```

Screen Shot 2020-01-15 at 3 37 39 PM

jtpio:
To do the upgrade, I created a new conda environment locally from scratch but haven't seen this.

jtpio:
And also:

```
rm -rf node_modules/
rm yarn.lock
```

afshin:
The `git clean -fdx` removes those files. I will try in a new conda environment to check again.

afshin:
It works in a brand new conda environment. I wonder what my issue is.

jtpio:
Good question.

We could try to reproduce a "manual upgrade" to see if that is worth the effort.

afshin:
The issue came from `ipywidgets`. The old version of `@jupyter-widgets/jupyterlab-manager` was still installed and the JupyterLab 2 beta accepted it but failed at run-time. I think this issue was fixed in a recent PR Steve did, but I'm not 100% sure of that. (cc: @blink1073 @jasongrout)

This fixed it:
Screen Shot 2020-01-15 at 4 00 32 PM

afshin:
This PR is included in the JupyterLab 1.x to 2.x migration guide: https://github.com/jupyterlab/jupyterlab/pull/7808

PR #343, opened 2020-01-16
Add keywords to package.json

jtpio:
To make the extension discoverable via the extension manager.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,11 @@
   "name": "@jupyterlab/debugger",
   "version": "0.2.0-alpha.1",
   "description": "JupyterLab - Debugger Extension",
+  "keywords": [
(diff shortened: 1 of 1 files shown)


afshin:
Oh man, the new libraries are making tests flaky. It seems likely that either we or the libraries have a race condition.

jtpio:
Yes, I thought this was fixed by https://github.com/jupyterlab/debugger/pull/337#issuecomment-574662927.

We should check one more time whether the kernels are properly awaited. The problem is that these issues are difficult to reproduce locally (tests usually always pass).

jtpio:
We've seen those failures in the past from time to time, but they seem to be more frequent after the switch to 2.0.

PR #345, opened 2020-01-17
Improve kernel restart handling

### LINKED ISSUE #322: Better way to handle kernel restarts

jtpio:
Fixes #322.

Follow-up on #337. 

This should improve the kernel restart logic:

kernel-restarts-3

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -13,7 +13,7 @@ import { FileEditor } from '@jupyterlab/fileeditor';
 
 import { NotebookPanel } from '@jupyterlab/notebook';
 
-import { Session } from '@jupyterlab/services';
+import { Kernel, Session } from '@jupyterlab/services';
 
 import { DebuggerModel } from './model';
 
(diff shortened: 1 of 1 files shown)


jtpio:
The test failure for this one looks odd:

```
TypeError: Cannot read property 'sendRequest' of undefined

      172 |   describe('#debugInfo', () => {
      173 |     it('should return the state of the current debug session', async () => {
    > 174 |       const reply = await debugSession.sendRequest('debugInfo', {});
          |                                        ^
      175 |       expect(reply.body.isStarted).to.be.true;
      176 | 
      177 |       const breakpoints = reply.body.breakpoints;
```

`debugSession` should be defined in the `beforeEach` hook here: https://github.com/jupyterlab/debugger/blob/b267d4a7b96678bfdc3771a6182239987aa18629/tests/src/session.spec.ts#L126

Issue #177, opened 2019-11-13
Display nested data structures


jtpio:
Now that #125 has been merged, we can request more variables when expanding a tree node.

However we don't support nested data structures yet, such as a `dict` in a `dict`:

![image](https://user-images.githubusercontent.com/591645/68774630-afbe4a00-062d-11ea-899e-3e75622cef78.png)

But it looks like we could do some further refactoring, and end up having:

- The current scopes and variables, as returned by the DAP, stored in memory in the model
- A mapper that transforms the data from the DAP, and maintains a JSON-like data structure that represents the variables as the state
- The react view that consumes the JSON-like state and renders the tree as is
- The tree view component should also support events when a node is expanded
- On expand, a request is sent to the kernel to retrieve more information for the variable using the `variableReference`

There is already some of this logic in the current implementation, but we should try to separate the state management, rendering and event handling even more.

Also we should check whether or not the current `variableExpanded` signal is needed in the model. Or if we can make the request for more variables via the service directly:

https://github.com/jupyterlab/debugger/blob/8192e8a4185d2f4d8f77b937cd10ba752d7b0531/src/variables/index.ts#L82-L84

jtpio:
At the moment, there is a short period of time where we see the previous content:

![image](https://user-images.githubusercontent.com/591645/68784856-89a0a600-063d-11ea-85cf-670bbf0f688b.png)

As a workaround we could check if it's possible to override the empty dict when the node is collapsed.

This was introduced in https://github.com/jupyterlab/debugger/pull/125/commits/dbead195a59b24f12c80fc43db2960939ee650f3

KsavinN:
#347

Issue #272, opened 2019-12-11
[Variable Inspector] Switch back to table view?


jtpio:
I was showing the state of the debugger to someone currently using Spyder instead of Jupyter notebooks. The feedback was that the tree view didn't feel as intuitive as the table view in Spyder to inspect variables.

Maybe we should switch back to a table view instead of a tree view? For reference this is what it looks like in Spyder:

![spyder-variable-explorer](https://user-images.githubusercontent.com/591645/70617111-27bd6700-1c10-11ea-9c2a-177753c0cbc2.gif)

This could have some benefits:

- The table view is a more familiar interface for users that are not necessarily developers
- Double-clicking on a variable (for example when inspecting a nested dictionary or a numpy array) could create a new table with the new content. A "back arrow" would bring back to the top-level view  
- For widgets: hovering on the value in the table would bring a popup with the rendered (and interactive) widget
- It would also be easier to detect when a new `variables` debug request must be sent to the kernel

afshin:
I think this is a good idea, because as we talked about on yesterday's call, our "tree" really is only one level deep anyway. The main benefit of the tree is collapsing the `local` vs. `global` variables, but perhaps that can just be two tables.

jtpio:
Two tables sounds good (or one table per `scope` to stick to the DAP).

Also this would make it convenient to edit variables on the fly, for example by double-clicking on a cell in the table, and using the [`evaluate` request from the DAP](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Evaluate) to send the new value.

dhirschfeld:
> *Double-clicking on a variable (for example when inspecting a nested dictionary or a numpy array) could create a new table with the new content. A "back arrow" would bring back to the top-level view*

...or the popup could have the tree-view of the object. Not everything maps well to a 2D table and for custom classes I really like the ability to expand the object to introspect objects that aren't in the repr

dhirschfeld:
For ref, the variable inspector in PyCharm is a tree-view which I quite like - I guess it's what you're used to

![image](https://user-images.githubusercontent.com/881019/70667798-8856ad80-1cbd-11ea-970e-6905edfbb6e3.png)

jtpio:
Thanks @dhirschfeld. A tree can be more natural for a structure like a `class` or a `dict`.

It looks like PyCharm also has a `View As Array` that opens a `numpy` array as a table.

afshin:
[...]

PR #347, opened 2020-01-17
Switch to Table View

KsavinN:
related with #272 

table_tree


dhirschfeld:
Just wondering if the viewer windows shouldn't be ephemeral by default - i.e. after opening a viewer, if the user then clicks on the background the viewer window will disappear (be disposed) automatically. If the user wants the viewer window to remain on top they can pin it by clicking on a "pin" toggle button.

jtpio:
Also wondering how it plays when a new widget (for example another notebook) is open in the main area.

KsavinN:
@jtpio @dhirschfeld  What do you think about this, maybe details table should appear as like normal widget in mainArea?
newWay

dhirschfeld:
FWIW PyCharm allows you to choose/switch between floating or docked for their widgets which I guess gives the *maximum* flexibility/configurability. However I think this is pretty good as is - it gives the user the ability to arrange the panels however they want.

jtpio:
This looks good, thanks @KsavinN.

It looks like it feels more natural than the floating windows for now, and more integrated to JupyterLab.

KsavinN:
Hi, I implement our new Tree view and remove React-inspector. This should resolve #177.
neewTree

jtpio:
Nice, this has the advantage of being able to fully control click / expand events.

What does it look like with other types of structures, such as classes and numpy arrays?

KsavinN:
> Nice, this has the advantage of being able to fully control click / expand events.
> 
> What does it look like with other types of structures, such as classes and numpy arrays?

with class: 
class


with numpy - it's look like put a lot information from backend:
numpy

jtpio:
Yes for numpy arrays (and potentially more advanced data structures) we should check the `variables` response to see if it makes sense to keep everything or not.

src/variables/table.tsx
```diff
@@ -0,0 +1,191 @@
...
+      variableReference: variable.variablesReference,
+      title: variable.evaluateName
+    });
+  };
+
+  const Tbody = (variables: VariablesModel.IVariable[]) => (
+    <tbody>
+      {variables?.map(variable => (
+        <tr
+          onClick={() => onClickVariable(variable)}
```
jtpio: We'll probably want a double click here to avoid opening too many panels.

src/index.ts
```diff
@@ -49,6 +57,8 @@ export namespace CommandIDs {
   export const stepIn = 'debugger:stepIn';
 
   export const stepOut = 'debugger:stepOut';
+
+  export const addDetials = 'debugger:variableDetials';
```
```suggestion
  export const addDetails = 'debugger:variableDetails';
```

jtpio:
Also we probably want to adjust a couple of CSS variables to support the dark theme:

image

src/index.ts
```diff
@@ -244,6 +254,63 @@ const tracker: JupyterFrontEndPlugin<void> = {
...
+/*
+ * A plugin for add widget with detial of veriable.
+ */
+const details: JupyterFrontEndPlugin<void> = {
+  id: '@jupyterlab/debugger:details',
+  autoStart: true,
+  requires: [IDebugger],
+  activate: (app: JupyterFrontEnd, service: IDebugger) => {
+    const { commands, shell } = app;
+    const detailsTracker = new WidgetTracker<MainAreaWidget<VariableDetails>>({
```
jtpio: Should `detailsTracker` be called `tracker` if that's the only tracker in this plugin?

src/variables/index.ts
```diff
@@ -45,10 +83,37 @@ export class Variables extends Panel {
...
+    case 'bool':
+      return value;
+    case 'str':
+      return value.slice(1, value.length - 1);
+    case 'list':
+      return 'List';
+    case 'dict':
+      return 'Dict';
+    case 'function':
+      return 'Function';
```
jtpio: Maybe we could just keep `list`, `dict` and `function` as-is?

src/variables/index.ts
```diff
@@ -61,5 +126,11 @@ export namespace Variables {
      * The variables model.
      */
     model: VariablesModel;
+    /**
+     * The debugger service.
+     */
+    service: IDebugger;
+
+    commands: CommandRegistry;
```
jtpio: Missing docstring.

src/variables/model.ts
```diff
@@ -24,6 +24,15 @@ export class VariablesModel {
     this._changed.emit();
   }
 
+  set details(variables: VariablesModel.IVariable[]) {
+    console.log({ variables });
```
jtpio: It looks like this console.log can be removed.

src/variables/model.ts
```diff
@@ -38,6 +47,10 @@ export class VariablesModel {
     return this._variableExpanded;
   }
 
+  get variableCliked(): ISignal<this, VariablesModel.IVariable> {
```
jtpio: Is this signal used?

src/variables/model.ts
```diff
@@ -46,9 +59,15 @@ export class VariablesModel {
     this._variableExpanded.emit(variable);
   }
 
+  variableGetDetails(variable: VariablesModel.IVariable) {
```
jtpio: Is this method used?

src/variables/table.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { convertType } from '.';
+
+import { IDebugger } from '../tokens';
+
+import { CommandIDs } from '..';
+
+import { CommandRegistry } from '@lumino/commands';
+
+/**
+ * The body for Table Variables .
```
jtpio:
```suggestion
 * The body for a tables of Variables.
```

src/variables/table.tsx
```diff
@@ -0,0 +1,166 @@
...
+export class VariablesBodyTable extends ReactWidget {
+  /**
+   * Instantiate a new Body for the Table Variables .
+   * @param model The model for the variables.
+   */
+  constructor(options: VariablesBodyTable.IOptions) {
+    super();
+    this._model = options.model;
+    this._commands = options.commands;
+    this.addClass('jp-DebuggerVariables-body');
```
jtpio: Should this call to `addClass` be placed at the end of the constructor, to follow the same pattern as for the other components?

src/variables/table.tsx
```diff
@@ -0,0 +1,166 @@
...
+    this._model.changed.connect(this.updateScopes, this);
+  }
+
+  private updateScopes(self: VariablesModel) {
```
jtpio: How about replacing `self` by a more descriptive name?

jtpio: Also this method could be renamed to `_updateScopes` to follow the code style, and have a docstring.

src/variables/table.tsx
```diff
@@ -0,0 +1,166 @@
...
+      </>
+    );
+  }
+
+  private _model: VariablesModel;
+  private _scopes: VariablesModel.IScope[] = [];
+  private _commands: CommandRegistry;
+}
+
+export class VariableDetails extends ReactWidget {
```
jtpio: Let's try to add docstrings to the new classes, methods, and attributes.

src/variables/table.tsx
```diff
@@ -0,0 +1,166 @@
...
+  constructor(options: VariablesDetails.IOptions) {
+    super();
+    const { details, commands, model, service, title } = options;
+    this._variables = details;
+    this._commands = commands;
+    this._model = model;
+    this._model.changed.connect(this.updateScopes, this);
+    this.title.iconClass = 'jp-VariableIcon';
+    this.addClass('jp-debugger-details');
```
jtpio: Same comment as above for `addClass`.

src/variables/table.tsx
```diff
@@ -0,0 +1,166 @@
...
+  commands: CommandRegistry;
+}) => {
+  const [variables, setVariables] = useState(data);
+  const [selected, setSelected] = useState(null);
+
+  useEffect(() => {
+    setVariables(data);
+  }, [data]);
+
+  const onClickVariable = async (variable: VariablesModel.IVariable) => {
```
jtpio:
```suggestion
  const onVariableClicked = async (variable: VariablesModel.IVariable) => {
```

src/variables/tree.tsx
```diff
@@ -0,0 +1,159 @@
...
+import { ReactWidget } from '@jupyterlab/apputils';
+
+import React, { useEffect, useState } from 'react';
+
+import { IDebugger } from '../tokens';
+
+import { VariablesModel } from './model';
+
+/**
+ * The body for Tree Variables .
```
jtpio:
```suggestion
 * The body for Tree Variables.
```

src/variables/tree.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  const styleName = {
+    color: 'var(--jp-mirror-editor-attribute-color)'
+  };
+
+  const styleType = {
+    color: 'var(--jp-mirror-editor-string-color)'
+  };
+
+  const nonExpanded =
```
jtpio: How about renaming this variable to something like `expandable`, and invert its value?

src/variables/tree.tsx
```diff
@@ -0,0 +1,159 @@
...
+    variable.variablesReference === 0 && variable.type !== 'function';
+
+  const onClickVariable = async (e: React.MouseEvent) => {
+    if (nonExpanded) {
+      return;
+    }
+    e.stopPropagation();
+    const variableDetials = await service.getVariableDetails(
```
jtpio:
```suggestion
    const variableDetails = await service.getVariableDetails(
```

style/variables.css
```diff
@@ -6,5 +6,96 @@
...
+  font-size: 12px;
+  margin-right: 3px;
+  user-select: none;
+}
+
+.jp-DebuggerVariables header .jp-Toolbar {
+  margin-left: auto;
+}
+
+.jp-debugger-details {
```
jtpio: Maybe these could follow the same convention as the other CSS classes.

jtpio:
In a next iteration we could try to align the different entries in the tree or add some padding:

image

src/variables/table.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  const Tbody = (variables: VariablesModel.IVariable[]) => (
+    <tbody>
+      {variables?.map(variable => (
+        <tr
+          onDoubleClick={() => onVariableClicked(variable)}
+          key={variable.evaluateName}
+        >
+          <td>{variable.name}</td>
+          <td>{convertType(variable)}</td>
```
jtpio: Shouldn't that be `variable.type` instead? Otherwise this will print `1` for `x = 1` for example.

src/variables/table.tsx
```diff
@@ -0,0 +1,216 @@
...
+      variableReference: variable.variablesReference,
+      title: variable.evaluateName
+    });
+  };
+
+  const Tbody = (variables: VariablesModel.IVariable[]) => (
+    <tbody>
+      {variables?.map(variable => (
+        <tr
+          onDoubleClick={() => onVariableClicked(variable)}
```
jtpio: It would be nice if we could still keep the single click to highlight the entire row.

src/variables/table.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  const onVariableClicked = (variable: VariablesModel.IVariable) => {
+    if (selected === variable) {
+      setSelected(null);
+      return;
+    }
+    setSelected(variable);
+  };
+
+  const onVariableDbClicked = async (variable: VariablesModel.IVariable) => {
```
jtpio:
```suggestion
  const onVariableDoubleClicked = async (variable: VariablesModel.IVariable) => {
```

src/variables/tree.tsx
```diff
@@ -0,0 +1,177 @@
...
+  };
+
+  const styleType = {
+    color: 'var(--jp-mirror-editor-string-color)'
+  };
+
+  const expandable =
+    variable.variablesReference !== 0 || variable.type === 'function';
+
+  const onClickVariable = async (e: React.MouseEvent) => {
```
jtpio: How about naming it `onVariableClicked` for consistency?

src/variables/table.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  useEffect(() => {
+    setVariables(data);
+  }, [data]);
+
+  const onVariableClicked = (variable: VariablesModel.IVariable) => {
+    if (selected === variable) {
+      setSelected(null);
+      return;
+    }
```
jtpio: This will remove the selection when clicking on the same variable again. It feels like the selection should not be toggled here.

src/variables/index.ts
```diff
@@ -45,10 +83,37 @@ export class Variables extends Panel {
...
+    case 'str':
+      return value.slice(1, value.length - 1);
+    case 'list':
+      return type;
+    case 'dict':
+      return type;
+    case 'function':
+      return type;
+    default:
+      return value;
```
jtpio: Why make `list`, `dict` and `function` special, and not keep them in the `default`?

```typescript
default:
  return type;
```

There could be plenty of other types.

src/index.ts
```diff
@@ -244,6 +254,61 @@ const tracker: JupyterFrontEndPlugin<void> = {
...
+        const widget = new MainAreaWidget<VariableDetails>({
+          content: new VariableDetails({
+            commands,
+            service,
+            details,
+            model,
+            title
+          })
+        });
+        widget.id = `jp-debugger-details-${title}`;
```
jtpio: This refers to a CSS class that does not exist anymore?

KsavinN: No, that's id needed for adding to shell and for tracking. Do you think we should use a different method of creating id for these widgets?

jtpio: Ah ok I thought there was a link to the previous CSS class (see https://github.com/jupyterlab/debugger/pull/347#discussion_r373555799).

KsavinN: Yeah, that id is similar for last css style.

Issue #339, opened 2020-01-15
No session name in the sidebar on page refresh


jtpio:
The name of the session was added to the top of the sidebar in #331.

However it looks like it is not set when the page is refreshed, and then the debugger enabled:

![session-name-refresh](https://user-images.githubusercontent.com/591645/72426117-71f2b480-3789-11ea-8c60-45f1427f1237.gif)

jtpio:
Fixed by https://github.com/jupyterlab/debugger/pull/348.

PR #348, opened 2020-01-17
No header name bug

KsavinN:
related with #339 #351


src/session.ts
```diff
@@ -62,7 +62,9 @@ export class DebugSession implements IDebugger.ISession {
     }
 
     this._connection.iopubMessage.connect(this._handleEvent, this);
-    this._connectionChanged.emit(this._connection);
+    this._connectionChanged.emit(connection);
+    // put setTimeout to resolve issue with empty header after refresh page
+    setTimeout(() => this._connectionChanged.emit(connection));
```
afshin: Does this work?
```typescript
requestAnimationFrame(() => {
  this._connectionChanged.emit(connection);
}); 
```

KsavinN: request work too

KsavinN:
Found some new approach for this issue

KsavinN:
I found and resolve the reason why test don't go succeed and add super.dispose :)

jtpio:
Nice. Indeed it could have been that the editor handler was not instantiated (no source displayed).

src/model.ts
```diff
@@ -77,6 +77,31 @@ export class DebuggerModel implements IDebugger.IModel {
...
+  }
+
+  /**
+   * Set the current debugger title.
+   */
+  set title(title: string) {
+    if (title === this._title) {
+      return;
+    }
+    this._title = title || '-';
```
jtpio: Should we use `??` instead of `||`?

src/session.ts
```diff
@@ -62,7 +62,6 @@ export class DebugSession implements IDebugger.ISession {
     }
 
     this._connection.iopubMessage.connect(this._handleEvent, this);
-    this._connectionChanged.emit(this._connection);
```
jtpio: Is this signal emitted somewhere else? Or does that mean it could be removed from the interfaces as well?

src/sources/body.ts
```diff
@@ -69,8 +69,9 @@ export class SourcesBody extends Widget {
     if (this.isDisposed) {
       return;
     }
-    this._editorHandler.dispose();
```
jtpio: Good catch :+1:

src/session.ts
afshin: It doesn't look like it's used anywhere, so this PR removes it from the interface in `token.ts` as well.

jtpio: Yes, it was just rebased to remove it.

afshin: Gotcha, I couldn't figure out why I couldn't find it 😂

src/service.ts
```diff
@@ -242,10 +242,14 @@ export class DebuggerService implements IDebugger, IDisposable {
...
 
+    if (this.isStarted || autoStart) {
+      this._model.title = this.isStarted ? this.session?.connection?.name : '-';
+    }
+
     this._model.breakpoints.restoreBreakpoints(bpMap);
     if (stoppedThreads.size !== 0) {
       await this._getAllFrames();
-    } else {
+    } else if (this.isStarted) {
```
jtpio: Should we leave this change out of this PR for now?

I believe this is to handle #351, but it looks like other changes would be required too.

jtpio:
Testing locally, it appears that showing the state of the other debug session in the notebook to fix #351 leads to other issues such as:

sidebar-state-sessions

How about tackling that in a different PR, so this one could be merged? (otherwise the rest looks good)

KsavinN:
> Testing locally, it appears that showing the state of the other debug session in the notebook to fix #351 leads to other issues such as:
> 
> sidebar-state-sessions
> 
> How about tackling that in a different PR, so this one could be merged? (otherwise the rest looks good)

Good catch. I will tried resolve this today. If i it will take more time and code we can just merge that and resolve this issue in another PR

KsavinN:
Ok. it will take more code (maybe not more time). Let's merge That :)

jtpio:
How about reverting this first?

https://github.com/jupyterlab/debugger/pull/348#discussion_r374071851

KsavinN:
I reverted. For now when user change to untoggled widget it will clear model and set Tittel to '-'.

src/service.ts
```diff
@@ -242,10 +242,15 @@ export class DebuggerService implements IDebugger, IDisposable {
...
 
+    if (this.isStarted || autoStart) {
+      this._model.title = this.isStarted ? this.session?.connection?.name : '-';
+    }
+
     this._model.breakpoints.restoreBreakpoints(bpMap);
     if (stoppedThreads.size !== 0) {
       await this._getAllFrames();
     } else {
+      // this._model.title = this.isStarted ? this.session?.connection?.name : '-';
```
jtpio: Let's remove the comment.

jtpio:
> I reverted. For now when user change to untoggled widget it will clear model and set title to '-'.

Sounds good.

PR #350, opened 2020-01-21
Attempt at fixing the tests on CI

jtpio:
This is an attempt at fixing the tests on CI, by using the same pattern as in core JupyterLab when it comes to initializing the session context and waiting for the kernel to be ready.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -102,7 +102,6 @@ export class DebugSession implements IDebugger.ISession {
    * Start a new debug session
    */
   async start(): Promise<void> {
-    await this._ready();
     await this.sendRequest('initialize', {
       clientID: 'jupyterlab',
       clientName: 'JupyterLab',
@@ -125,7 +124,6 @@ export class DebugSession implements IDebugger.ISession {
    * Stop the running debug session.
    */
   async stop(): Promise<void> {
-    await this._ready();
     await this.sendRequest('disconnect', {
       restart: false,
       terminateDebuggee: true
@@ -137,7 +135,6 @@ export class DebugSession implements IDebugger.ISession {
    * Restore the state of a debug session.
    */
   async restoreState(): Promise<IDebugger.ISession.Response['debugInfo']> {
-    await this._ready();
     const message = await this.sendRequest('debugInfo', {});
     this._isStarted = message.body.isStarted;
     return message;
--- a/tests/src/service.spec.ts
+++ b/tests/src/service.spec.ts
@@ -1,10 +1,10 @@
 import { expect } from 'chai';
 
-import { ISessionContext } from '@jupyterlab/apputils';
+import { Session } from '@jupyterlab/services';
 
-import { createSessionContext } from '@jupyterlab/testutils';
+import { createSession } from '@jupyterlab/testutils';
 
-import { PromiseDelegate } from '@lumino/coreutils';
+import { PromiseDelegate, UUID } from '@lumino/coreutils';
 
 import { DebuggerModel } from '../../lib/model';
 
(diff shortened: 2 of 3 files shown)


jtpio:
This needs multiple re-runs to check how it behaves.

jtpio:
Also failing after adding an artificial `await sleep(2000)`:

https://dev.azure.com/jupyterlab/jupyterlab/_build/results?buildId=5016

```diff
diff --git a/tests/src/service.spec.ts b/tests/src/service.spec.ts
index 288dfb0..53666d8 100644
--- a/tests/src/service.spec.ts
+++ b/tests/src/service.spec.ts
@@ -2,7 +2,7 @@ import { expect } from 'chai';

 import { ISessionContext } from '@jupyterlab/apputils';

-import { createSessionContext } from '@jupyterlab/testutils';
+import { createSessionContext, sleep } from '@jupyterlab/testutils';

 import { PromiseDelegate } from '@lumino/coreutils';

@@ -35,6 +35,7 @@ describe('Debugging support', () => {
       xpython.session?.kernel?.info,
       ipykernel.session?.kernel?.info
     ]);
+    await sleep(2000);
   });

   afterAll(async () => {
diff --git a/tests/src/session.spec.ts b/tests/src/session.spec.ts
index 43c50bf..83c25c5 100644
--- a/tests/src/session.spec.ts
+++ b/tests/src/session.spec.ts
@@ -5,7 +5,7 @@ import { expect } from 'chai';

 import { ISessionContext } from '@jupyterlab/apputils';

-import { createSessionContext } from '@jupyterlab/testutils';
+import { createSessionContext, sleep } from '@jupyterlab/testutils';

 import { find } from '@lumino/algorithm';

@@ -28,6 +28,7 @@ describe('DebugSession', () => {
     });
     await sessionContext.initialize();
     await sessionContext.session?.kernel?.info;
+    await sleep(2000);
   });

   afterEach(async () => {
@@ -122,6 +123,8 @@ describe('protocol', () => {
     });
     await sessionContext.initialize();
     await sessionContext.session?.kernel?.info;
+    await sleep(2000);
+
     debugSession = new DebugSession({
       connection: sessionContext.session
     });
```

jtpio:
One main difference compared to before switching to 2.0 is the fact that the `DebugSession` now takes a `Session.ISessionConnection`, while we used to pass the `ClientSession` directly before:

2.0:

https://github.com/jupyterlab/debugger/blob/ce99123a5f01675aba300565d6532bcd0e075964/tests/src/service.spec.ts#L72

1.x:

https://github.com/jupyterlab/debugger/blob/db16526456a82b83768345fbcffb1b5cf68fb5f5/tests/src/service.spec.ts#L74

jtpio:
Since `DebugSession` takes a `Session.ISessionConnection`, we can try to create a session connection instead of a session context, like in this commit: https://github.com/jupyterlab/debugger/pull/350/commits/ae9a0188624329717a6b0bc90e8c90d3fa27ba9f

This run seems to be passing, but we would need a couple more to double check: https://dev.azure.com/jupyterlab/jupyterlab/_build/results?buildId=5041&view=logs&jobId=ebc2fedd-630a-59f1-5615-d345b5a141d6

jtpio:
cc @afshin if you also want to test locally

jtpio:
4 consecutive green, this is a higher rate than before:

image


I think the problem was that a race condition would make the `SessionContext` dispose its session connection on `sessionContext.dispose()`, while still being referenced in the `DebugSession`:

https://github.com/jupyterlab/jupyterlab/blob/a09e0ec2a98ecff254577786d87a086e5d070b08/packages/apputils/src/sessioncontext.tsx#L525-L527

Which could explain why we would often get: 

https://github.com/jupyterlab/debugger/blob/ce99123a5f01675aba300565d6532bcd0e075964/src/session.ts#L189-L194

src/session.ts
```diff
@@ -102,7 +102,6 @@ export class DebugSession implements IDebugger.ISession {
    * Start a new debug session
    */
   async start(): Promise<void> {
-    await this._ready();
```
jtpio: Those shouldn't be needed, as there is already a call to `this._ready()` in `sendRequest`.

PR #352, opened 2020-01-24
v0.2.0-alpha.2

afshin:


jtpio:
Should we also update the version number in `postBuild` when we cut a new release?

https://github.com/jupyterlab/debugger/blob/881309843faae769a044adeb717dcc708b94ef26/binder/postBuild#L1

afshin:
Yes that makes sense! :+1:

PR #357, opened 2020-02-04
Toggle life bug

### LINKED ISSUE #346: Toggle button not shown when switching from "No Kernel"
As mentioned in https://github.com/jupyterlab/debugger/pull/337#issuecomment-574604673, switching to "No Kernel" and to then back to xpython doesn't show the toggle button in the toolbar:

no-kernel-toggle-button

### LINKED ISSUE #356: Keep active session
image

After switching to another notebook, our model has a new session. It should use steel old session if new one doesn't have a toggle on the debugger

KsavinN:
Fixes #346
Fixes #356
 
its work
nokernel

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -3,9 +3,11 @@
 
 import { JupyterFrontEnd } from '@jupyterlab/application';
 
-import { ISessionContext, ToolbarButton } from '@jupyterlab/apputils';
+import { ToolbarButton } from '@jupyterlab/apputils';
 
 import { ConsolePanel } from '@jupyterlab/console';
+
+import { IChangedArgs } from '@jupyterlab/coreutils';
 
 import { DocumentWidget } from '@jupyterlab/docregistry';
 
@@ -25,8 +27,6 @@ import { ConsoleHandler } from './handlers/console';
 
 import { FileHandler } from './handlers/file';
 
-import { IChangedArgs } from '@jupyterlab/coreutils';
-import { IKernelConnection } from '@jupyterlab/services/lib/kernel/kernel';
 import { NotebookHandler } from './handlers/notebook';
 
 /**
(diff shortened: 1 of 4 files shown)


src/handler.ts
```diff
@@ -87,36 +93,67 @@ export class DebuggerHandler {
...
   ): Promise<void> {
-    const updateHandler = async () => {
+    if (!connection) {
+      delete this._kernelChangedHandlers[widget.id];
+      delete this._statusChangedHandlers[widget.id];
       return this._update(widget, connection);
+    }
+
+    const kernelChanged = () => {
+      void this._update(widget, connection);
```
jtpio: Small thing here: any reason for not returning directly?

```suggestion
      return this._update(widget, connection);
```

src/handler.ts
```diff
@@ -87,36 +93,67 @@ export class DebuggerHandler {
...
 
-    // setup handler when the kernel changes
-    const kernelChangedHandler = this._kernelChangedHandlers[connection.path];
     if (kernelChangedHandler) {
       connection.kernelChanged.disconnect(kernelChangedHandler);
     }
-    connection.kernelChanged.connect(updateHandler);
-    this._kernelChangedHandlers[connection.path] = updateHandler;
+    this._kernelChangedHandlers[widget.id] = kernelChanged;
+    connection.kernelChanged.connect(kernelChanged);
```
jtpio: Do we need to keep track of both `kernelChanged` signals?

The one from `SessionConnection` should be proxied by the `SessionContext`: https://github.com/jupyterlab/jupyterlab/blob/d3cc2470976d3ea285abc6f855ad193445fca9e7/packages/apputils/src/sessioncontext.tsx#L712

jtpio: The answer is most likely yes because of the files.

jtpio:
It looks like the handler would deserve one more pass of refactoring, as the logic is now getting more complicated.

src/handler.ts
```diff
@@ -262,10 +312,31 @@ export class DebuggerHandler {
   private _type: DebuggerHandler.SessionType;
   private _shell: JupyterFrontEnd.IShell;
   private _service: IDebugger;
+  private _previousConnection: Session.ISessionConnection;
```
jtpio: Not sure having this `_previousConnection` is the way to go, as it adds another piece of information to keep track of.

Maybe an alternative would be to have something like `service.isAvailable()` but to ask whether the debug session has started? (using the `debugInfo` request).

jtpio:
As discussed, this seems to be fixing #346 and #356.

I'll opened a new issue to track the refactoring of the debug handler.

PR #360, opened 2020-02-10
Add a variable filter to the settings

### LINKED ISSUE #179: [VariableInspector] Hide private attributes/methods by default
Displaying all private `_*` and dunder `__*__` attributes/methods adds a lot of clutter. It would be good to collect these under top level `_` and `__` keys so users can  more easily focus on what is important/relevant.

A case in point is ipython where the tab completion doesn't show any attributes/methods starting with an underscore unless the user explicitly types `_` before requesting completions.

jtpio:
Fixes #179. 

Add a list to the debugger settings to be able to hide variables from the UI by default.

The current list of hidden variables is as follows:

```json
[
  "display",
  "ptvsd",
  "__annotations__",
  "__builtins__",
  "__doc__",
  "__loader__",
  "__name__",
  "__package__",
  "__spec__"
]
```

Until now, these variables would show in the list of variables, but could lead to some confusion as some of them are related to the debugger implementation details (`ptvsd`).

This list can be seen as an equivalent to gitignore files, where users can choose which variables to hide and show.

In follow-up PRs, we can add a show / hide button to the Variables toolbar to toggle the variables directly. 

### TODO

- [x] Add variable filter to the settings
- [x] Filter variables from the tree and table

filter-variables

### DIFF
--- a/src/variables/tree.tsx
+++ b/src/variables/tree.tsx
@@ -42,10 +42,19 @@ export class VariablesBodyTree extends ReactWidget {
             key={scope.name}
             service={this._service}
             data={scope.variables}
+            filter={this._filter}
           />
         ))}
       </>
     );
+  }
+
+  /**
+   * Set the variable filter list.
+   */
+  set filter(filter: Set<string>) {
+    this._filter = filter;
+    this.update();
   }
 
   /**
@@ -61,6 +70,7 @@ export class VariablesBodyTree extends ReactWidget {
   }
 
   private _scopes: VariablesModel.IScope[] = [];
+  private _filter = new Set<string>();
   private _service: IDebugger;
 }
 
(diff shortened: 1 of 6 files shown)


martinRenou:
The fact that it's a setting, does it mean that it is configurable by the user?

jtpio:
Yes. That way, users could still show these variables if they want to.

We can also decided whether it's an opt-in or opt-out.

jtpio:
For example this could be used to filter out the regular notebook imports used for pandas, numpy, matplotlib and so on by adding the following to the list:

```json
["np", "pd", "plt"]
```

src/variables/tree.tsx
```diff
@@ -42,12 +42,21 @@ export class VariablesBodyTree extends ReactWidget {
...
     );
   }
 
+  /**
+   * Set the variable filter list.
+   */
+  set filter(filter: Set<string>) {
+    this._filter = filter;
+    this.update();
+  }
```
jtpio: The table and tree viewers could benefit from some refactoring, to avoid having this logic at 2 different places.

jtpio:
cc @afshin I think this should help give a clearer idea of why having a user defined filter list can be useful (based on what we discussed last time) (cf screencast in [the first comment](https://github.com/jupyterlab/debugger/pull/360#issue-373068082)).

jtpio:
Rebased onto the latest master.

@KsavinN were you able to try this PR locally?

jtpio:
A good point raised by @afshin is to actually have a mapping from kernel ids to a list of variables. For example:

```json
'xpython': [
  "display",
  "ptvsd",
  "__annotations__",
  "__builtins__",
  "__doc__",
  "__loader__",
  "__name__",
  "__package__",
  "__spec__"
]
```

This will let other kernels specify their own list of variables to filter.

PR #364, opened 2020-02-14
Fix toggle button for JLab 2.0rc0

### LINKED ISSUE #361: Toggle switch button with JLab 2.0.0rc0
After updating the local setup to JupyterLab 2.0.0rc0, there seems to be an issue with the toggle button:

icon-rc0

This is probably a good time to look into using the new `LabIcon` system, as mentioned in https://github.com/jupyterlab/debugger/pull/337#discussion_r365890511.

jtpio:
Fixes #361. 

This pretty much reverts the change for the toggle button CSS from https://github.com/jupyterlab/debugger/pull/337.

Relying on the inner `svg` element was not a robust solution anyway in the long run.

This seems to fix the issue with `2.0.0rc0`:

toggle-button-rc-fix

### DIFF
--- a/style/icons.css
+++ b/style/icons.css
@@ -109,7 +109,8 @@
   width: 35px;
 }
 
-.jp-Toolbar-item .jp-ToggleSwitch svg {
+.jp-Toolbar-item .jp-ToggleSwitch::before {
+  content: '';
   height: 10px;
   width: 10px;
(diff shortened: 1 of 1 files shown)


jtpio:
Tested with the latest rc2 and it shows up as expected:

image

Issue #366, opened 2020-02-21
Fixed postion of header of table during scroll


KsavinN:
![noSteekyheader](https://user-images.githubusercontent.com/34033913/75042388-9b0c1080-54be-11ea-94b0-7f8148833427.gif)

PR #369, opened 2020-02-24
Column re-sizable in Table Variable

KsavinN:
resolve this issue #366 #74.

Switch from table selectors to flex-box.

table

### DIFF
--- a/src/variables/index.ts
+++ b/src/variables/index.ts
@@ -65,8 +65,8 @@ export class Variables extends Panel {
   }
 
   private _header: VariablesHeader;
-  private _tree: Widget;
-  private _table: Widget;
+  private _tree: VariablesBodyTree;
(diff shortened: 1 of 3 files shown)


jtpio:
@KsavinN do you have a screencast to see what it looks like?

jtpio:
it looks like we should be able to use the data grid (`@lumino/datagrid`) to implement the variable table viewer.

This would have several benefits such as handling large numbers of variables (some users will have hundreds / thousands of variables for sure), resizing, and having other variables (data frames, numpy arrays) look the same everywhere when inspected.

The API is quite different though. I'll look into this in a separate PR.

KsavinN:
Cool, if you implement that I will close that PR without merge.

KsavinN:
Due to end up with implementing dataGrid #391 I close this PR without merge

PR #370, opened 2020-02-27
Make Clickable path for open the source

### LINKED ISSUE #267: Clicking on the path should open the source
*Originally suggested by @SylvainCorlay*

Now that we show the path in the `SOURCE` header, clicking on it should also open the read-only editor (same as clicking on the icon).

image

KsavinN:
resolve #267 Clicking on the path should open the source


jtpio:
Thanks @KsavinN, will check it out later today.

PR #372, opened 2020-03-02
Update to JLab 2.0 final

jtpio:
Clean up the package versions and setup instructions after the 2.0 final release.


jtpio:
Merging, as this is more of a mechanical update than any change to the code.

PR #376, opened 2020-03-02
Fix icons for the dark theme

### LINKED ISSUE #375: Icons with the dark theme
See also: https://github.com/jupyterlab/debugger/issues/355

There are still a couple of places where "icons as css background" don't show up correctly with the dark theme: 

image

jtpio:
Fixes #375.

image

### TODO

- [x] Callstack icons
- [x] Debugger icon in the toolbar


jtpio:
Now sounds like a good time to migrate to the new `LabIcon`.

style/icons/step-into.svg
```diff
@@ -1,3 +1,5 @@
-<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path fill-rule="evenodd" clip-rule="evenodd" d="M7.99998 9.53198H8.54198L12.447 5.62698L11.386 4.56698L8.74898 7.17698L8.74898 0.999985H7.99998H7.25098L7.25098 7.17698L4.61398 4.56698L3.55298 5.62698L7.45798 9.53198H7.99998ZM9.95598 13.013C9.95598 14.1175 9.06055 15.013 7.95598 15.013C6.85141 15.013 5.95598 14.1175 5.95598 13.013C5.95598 11.9084 6.85141 11.013 7.95598 11.013C9.06055 11.013 9.95598 11.9084 9.95598 13.013Z" fill="#616161"/>
+<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
+	<g class="jp-icon3" fill="#616161">
```
jtpio: The `jp-icon3` is needed for the icons to be styled correctly based on the theme (using the correct fill color).

jtpio:
https://github.com/jupyterlab/debugger/pull/376/commits/31d8bb00304b30e97b84a2e599e69b7e9faa4078 should fix most of the icons. 

Still the debugger toolbar icon to fix.

jtpio:
https://github.com/jupyterlab/debugger/pull/376/commits/949b2802a1344e0dfa7de61ed0096160ea5f16eb adds the debugger icon and the button as two separate items in the toolbar.

As a side effect the hover style now applies only to the switch:

image

cc @telamonian do you know if there is a more canonical way of achieving this with the new `LabIcon` system?

Until now the toolbar component was defined with:

https://github.com/jupyterlab/debugger/blob/2ee7a2ee3b10529b711dfb285a5ac48f80b7118d/src/handler.ts#L38-L46

And styled with:

https://github.com/jupyterlab/debugger/blob/2ee7a2ee3b10529b711dfb285a5ac48f80b7118d/style/icons.css#L90-L122

Which made it possible to have both `jp-BugIcon` and the custom switch.

jtpio:
Let's make the bug icon clickable as well is we keep the two items separate.

jtpio:
https://github.com/jupyterlab/debugger/pull/376/commits/d7e58e4886b5fdadb929951a192f1af61e1a76e4 implements this:

bug-icon-switch

It kind of feels "odd" that two items of the toolbar do the same thing though.

jtpio:
Maybe @tgeorgeux has some input for a nice switch button UX that includes an icon? It looks like core JupyterLab doesn't have such component yet, so it might still be uncertain what we want to have really.

Otherwise I suggest handling this as a separate issue.

telamonian:
> cc @telamonian do you know if there is a more canonical way of achieving this with the new `LabIcon` system and the `bugIcon` from `ui-components`?

Short answer is no. I am currently thinking about ways to support multiple icons/text labels in a single LabButton (or something) component. Part of that is figuring out the actual use cases, so thanks for this.

It's a bit hacky, but I think you can get what you really want (a single outer `<button>` with two child icons side by side) like this:

```
const button = new ToolbarButton({ 
   className: classes('jp-DebuggerSwitchButton', 'jp-ToggleSwitch'),
   icon: bugIcon, 
   onClick, 
   tooltip: 'Enable / Disable Debugger' 
 }); 
```

So the bug icon would be inline svg, while the toggle switch would continue to be CSS. Getting the styling right might be a bit tricky, though

telamonian:
> Yes, this is basically how it is at the moment (except that is uses `iconClass` instead of `icon`).

The difference would be that both bug icon and toggle icon would be children of the same large button, like before, instead of having

> two items of the toolbar do the same thing though.

jtpio:
@afshin It looks like something like this works well actually:

bug-icon-switch-label

afshin:
Awesome. This looks great to me.

tgeorgeux:
@jtpio I am working on switches here: https://github.com/jupyterlab/jupyterlab/issues/8010

I think if the icon is being used as a toggle it makes more sense to color the icon to indicate it's on or off. The design pattern doesn't exist in Jupyterlab to have an 'icon and toggle' together. I

Hope this helps, I'll be in touch.

jtpio:
Thanks @tgeorgeux for opening this issue :+1:

PR #378, opened 2020-03-04
Handle switching between two active debug sessions

### LINKED ISSUE #377: Switching between notebooks doesn't update the sidebar
From the current master:

switch-notebook-no-update

jtpio:
Fixes #377.

### DIFF
--- a/src/callstack/model.ts
+++ b/src/callstack/model.ts
@@ -21,8 +21,12 @@ export class CallstackModel {
    */
   set frames(newFrames: CallstackModel.IFrame[]) {
     this._state = newFrames;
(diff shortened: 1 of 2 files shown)


jtpio:
It looks like only the Variable Tree doesn't update itself properly:

variable-tree-update

jtpio:
Actually the example is using two different kernels, but the variable names are the same (`a` and `b`).

The VariableComponent in the tree view was using the variable name as a key, and since it's the same for both sessions React didn't detect a change.

jtpio:
It works for the table viewer because the name, type and value fields are placed in different `<td>` elements.

PR #381, opened 2020-03-06
Carret component #version_1

### LINKED ISSUE #325: Collapse the sections in the Debugger sidebar
Similar to the extension manager, it would be useful to be able to collapse the different section in the debugger sidebar:

collapse-sections

KsavinN:
carret
This resolve #325 
This is first-way which work better with a split panel and need less code but its not work properly as a collapse style.


KsavinN:
After discussion, we decide used behaviour from #382

PR #382, opened 2020-03-06
Caret Component #2

### LINKED ISSUE #325: Collapse the sections in the Debugger sidebar

KsavinN:
collapse
resolve #325 

Another way for collapse in splitPanel at Debugger Sidebar.


jtpio:
This option looks nicer than #381 from a user's perspective :+1:

src/collapse.ts
```diff
@@ -0,0 +1,116 @@
...
+      setNextWidgetTop(widget, nextWidget);
+      if (!nextWidget?.node.classList.contains('hide')) {
+        break;
+      }
+    }
+  }
+
+  private moveOnOpenPanel(splitWidgets: Widget[], elements: HTMLDivElement[]) {
+    let index = 0;
+    const calcutaeDiffPixel = (h1: string, h2: string) =>
```
jtpio:
```suggestion
    const calculatePixelDiff = (h1: string, h2: string) =>
```

jtpio:
Thanks @KsavinN for working on this.

I gave this a quick spin locally, and it looks like we can end up in situations like the following when dragging the handles of the `SplitPanel`:

collapse-sidebar-glitch

Not sure how easy it would be to prevent this (haven't looked at the logic in details yet).

KrzysztofSikoraCodete:
Peek 2020-04-07 19-33

@jtpio I've catch second example of this problem. Looks more clearer what's going on.

KrzysztofSikoraCodete:
panels

Issue #384, opened 2020-03-08
I can't install xeus-python with conda.


BigaMachine:
> Collecting package metadata (current_repodata.json): done
> Solving environment: failed with initial frozen solve. Retrying with flexible solve.
> Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
> Collecting package metadata (repodata.json): done
> Solving environment: - failed with initial frozen solve. Retrying with flexible solve.
> 
> CondaError: KeyboardInterrupt
> 

It blocked permanently when I install xeus-python.
What should I do? Thanks! :)
And I have installed @jupyterlab/debugger successfully, but nothing changed..

envs: Ubuntu 19.10, Anaconda 2019.10, conda 4.7.12.

afshin:
Is this the command you ran to install it?
```
conda install -c conda-forget xeus-python ptvsd
```

It needs to be the `conda-forge` channel. And you also need `ptvsd`.

afshin:
(I edited the above because I had a typo.)

BigaMachine:
Thanks for your reply :)
The command described in README.md that `conda install -c conda-forge xeus-python=0.6.7 notebook>=6 ptvsd` actually create  a "=6" file in my home folder. (This confused me a lot..)
So my command is `conda install -c conda-forge xeus-python=0.6.7 ptvsd`, since I found notebook has been installed in version 6.0.1 . 
And then an error described above occured.   :(

afshin:
Okay, just drop the `=0.6.7` part and I think it should work for you. We need to update the README.

afshin:
I think the trouble is that when you add constraints to arguments they need to be wrapped in quotation marks.

BigaMachine:
I'm not sure whether it would be working, cause a slash is spinning round and round and round when I try to install xeus-python solely without any other constraint.`conda install -c conda-forge xeus-python`

afshin:
Could you try running:
```
conda install --verbose -c conda-forge xeus-python
```
And paste the entire output of the command here?

BigaMachine:
Sure, I run this and it exhausted all my cpu until I press `Control^C`, and get nothing.

```bash
(base) bgmc@tp:~$ conda install --verbose -c conda-forge xeus-python
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
[...]

PR #385, opened 2020-03-08
Wrap constraint arguments in `conda` command in quotation marks.

afshin:
cf. https://github.com/jupyterlab/debugger/issues/384

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,14 +25,14 @@ jupyter labextension install @jupyterlab/debugger
 A kernel with support for debugging is also required to be able to use the debugger:
 
 ```bash
-conda install -c conda-forge "xeus-python>=0.6.7" "notebook>=6" ptvsd
+conda install -c conda-forge xeus-python=0.6.12 notebook=6 ptvsd
 ```
 
 ## Development
 
 ```bash
 # Create a new conda environment
-conda create -n jupyterlab-debugger -c conda-forge nodejs "xeus-python>=0.6.7" ptvsd "jupyterlab=2"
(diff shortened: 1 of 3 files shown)


jtpio:
Are those quotation marks really needed?

Sometimes shells interpret `>` and `=` differently.

afshin:
As the code is before this PR:
```
conda install -c conda-forge xeus-python>=0.6.7 notebook>=6 ptvsd
```
- [x] `bash` succeeds
- [ ] `zsh` fails

After the PR:
```
conda install -c conda-forge "xeus-python>=0.6.7" "notebook>=6" ptvsd
```

- [x] `bash` succeeds
- [x] `zsh` succeeds

afshin:
I ended up having two files: `=0.6.7 ` and `=6` in my home directory 😆

jtpio:
I'm using zsh and it works without the quotes.

afshin:
Hm. This is what happens to me:

conda

jtpio:
Ah ok, thanks for the screencast.

This seems to be caused by the `>`. We could drop it and keep `notebook=6` instead.

afshin:
Yeah, it is definitely the `>`.

I also realized there were several releases of `xeus-python` since we last updated the readme file. So we could just do `notebook=6 xeus-python=0.6`

afshin:
Oh wait, that's wrong. Because if you already have `xeus-python` version `0.6.1` it would not update.

jtpio:
Sounds good :+1: 

And the azure pipelines scripts could follow the same pinning strategy.

jtpio:
> Oh wait, that's wrong. Because if you already have `xeus-python` version `0.6.1` it would not update.

Ah right. Then yes let's pin it everywhere.

PR #399, opened 2020-03-16
Add GitHub Actions workflow

jtpio:
Trying out a GitHub Actions to run the tests alongside Azure Pipelines.


jtpio:
cc @afshin we can run both GH Actions and Azure Pipelines for a while and see how it goes (like JupyterLab at the moment)

jtpio:
Alright, let's remove Azure Pipelines and keep GH Actions only in this PR too.

PR #415, opened 2020-04-02
Add new test for pagination variables

KrzysztofSikoraCodete:
Peek 2020-04-02 11-45

I've just added test for paging request. The test could pass If the filtered variables result were less than 2. Now we have more than 1 even though we use pagination.

I checked session configuration and supportsVariablePaging, has true.

Any suggestions?


afshin:
Great, I think this is a sufficient test case to work against on the kernel side. Could you please open an issue in https://github.com/jupyter-xeus/xeus-python/ and tag this issue there?

Cheers, @KrzysztofSikoraCodete!

jtpio:
Thanks @KrzysztofSikoraCodete for looking into this :+1:

JohanMabille:
https://github.com/jupyter-xeus/xeus-python/pull/246 should fix it.

jtpio:
Thanks @JohanMabille!

afshin:
Awesome, thank you @JohanMabille!

jtpio:
Just re-triggered the tests and it looks like the conda tests are now passing with the latest `xeus-python`.

However the wheel is giving some errors: https://github.com/jupyterlab/debugger/pull/415/checks?check_run_id=612276325

jtpio:
See https://github.com/jupyter-xeus/xeus-python-wheel/issues/54 for more context.

The JupyterLab jest tests patch the jupyter paths here: https://github.com/jupyterlab/jupyterlab/blob/36a8da4d6bffea0fd9992a4d88627cffb46fe6a0/jupyterlab/tests/test_app.py#L88-L93

With the data directory becoming something like `/tmp/tmpgrqb54bk/jupyter_data`.

Which is why we install the `xpython` spec explicitly in the tests here: https://github.com/jupyterlab/debugger/blob/247735d59c1fa862048c38d3c16b5d18509c542b/tests/run-test.py#L15-L25

jtpio:
As discussed, let's disable the wheel tests for now and re-enable them in a separate PR.

jtpio:
Thanks!

KrzysztofSikoraCodete:
Thanks for merge! Nice to heard that.

Issue #408, opened 2020-03-26
Trailing slash in the breakpoint list


jtpio:
The path to the breakpoints now show a trailing `/`:

![image](https://user-images.githubusercontent.com/591645/77629331-a46b1b80-6f49-11ea-9452-a5fa283cb5b5.png)

As a reference, the screencast in the README still shows the path correctly.

jtpio:
Fixed by #423

PR #423, opened 2020-04-15
Remove slash from end of breakpoint path

KrzysztofSikoraCodete:
screen

https://github.com/jupyterlab/debugger/issues/408
@afshin

### DIFF
--- a/src/breakpoints/body.tsx
+++ b/src/breakpoints/body.tsx
@@ -88,6 +88,13 @@ const BreakpointCellComponent = ({
   breakpoints: IDebugger.IBreakpoint[];
   model: BreakpointsModel;
 }) => {
+  const removeFirstCharIfSlash = (breakpoint: IDebugger.IBreakpoint) => {
+    breakpoint.source.path =
+      breakpoint.source.path[0] === '/'
+        ? breakpoint.source.path.slice(1)
+        : breakpoint.source.path;
+    return breakpoint;
+  };
   return (
     <>
       {breakpoints
@@ -97,7 +104,7 @@ const BreakpointCellComponent = ({
         .map((breakpoint: IDebugger.IBreakpoint, index) => (
           <BreakpointComponent
             key={breakpoint.source.path + index}
-            breakpoint={breakpoint}
+            breakpoint={removeFirstCharIfSlash(breakpoint)}
             model={model}
           />
         ))}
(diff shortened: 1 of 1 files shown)


jtpio:
Does this still support the truncation at the beginning of the string with the ellipsis (`...`), as added in https://github.com/jupyterlab/debugger/pull/334?

Or does it undo this change?

KrzysztofSikoraCodete:
@jtpio After my change, support of (...) was unfortunelty removed.
I dived into the subject (directional: rtl) and i found some solution. <bdi> </bdi> tags.

How it works now you can see below.

bidi

jtpio:
Test failure looks similar to a previous issue with the browser check running on GitHub Actions:

- https://github.com/jupyterlab/extension-cookiecutter-ts/pull/53
- https://github.com/jupyterlab/jupyterlab/pull/7813
- https://github.com/actions/virtual-environments/issues/307

src/breakpoints/body.tsx
```diff
@@ -88,6 +88,13 @@ const BreakpointCellComponent = ({
   breakpoints: IDebugger.IBreakpoint[];
   model: BreakpointsModel;
 }) => {
+  const removeFirstCharIfSlash = (breakpoint: IDebugger.IBreakpoint) => {
+    breakpoint.source.path =
+      breakpoint.source.path[0] === '/'
```
jtpio, line 93: It would be interesting to see how this behaves with Windows path, since they don't start with a leading `/`.

jtpio: And I guess we still we want to keep the leading `/` and not remove it.

KrzysztofSikoraCodete: Peek 2020-04-17 13-10
Peek 2020-04-17 12-17

Third solution. Now path is not changed.

jtpio:
Thanks @KrzysztofSikoraCodete, this looks like a good solution at the moment to handle this.

The leading `/` is most likely treated as punctuation, and is then moved to the "right" with the `rtl` css property.

Also we will probably revisit the breakpoint UI at some point too: https://github.com/jupyterlab/debugger/issues/171

jtpio:
Strangely enough, the browser check on CI passed on this more recent PR: https://github.com/jupyterlab/debugger/pull/426.

KrzysztofSikoraCodete:
@jtpio Nice to heard that. I found on stackover flow the similar problem with direction `rtl`.
The reason of moved "/" to right side is what as you mentioned.

afshin:
Sweet! Cheers @KrzysztofSikoraCodete!

Issue #414, opened 2020-04-01
Extra breakpoints created every character typed


jasongrout:
When I edit a line with a breakpoint, it seems to create new breakpoints for the edits as I type them in, rather than just when I run the cell. This is in JLab 2.1.0rc0 and:

```
xeus-python               0.6.13           py38ha0d09dd_1    conda-forge
```
```
% jupyter labextension list
JupyterLab v2.1.0rc0
Known labextensions:
   app dir: /var/folders/ln/5623mk393v77n5l3rk2x88h00000gp/T/tmpenv.jWlUymub/env/share/jupyter/lab
        @jupyter-widgets/jupyterlab-manager v2.0.0  enabled  OK
        @jupyterlab/debugger v0.2.0  enabled  OK
        @jupyterlab/toc v3.0.0  enabled  OK
        @lckr/jupyterlab_variableinspector v0.5.0  enabled  OK
```

Here, when I type "myvar", I am just pausing, and after every character I type a new breakpoint appears.
![breakpoints](https://user-images.githubusercontent.com/192614/78193309-12727f80-742f-11ea-823c-21ced9b8ab64.gif)

jtpio:
Thanks for the issue and demo.

Yes, the content of the cell needs to be dumped as a file on the backend to be able to set breakpoints (more info in https://github.com/jupyterlab/debugger/issues/49, https://github.com/jupyterlab/debugger/issues/162, https://github.com/jupyterlab/debugger/issues/64)

Since the cells are identified using the murmur hash of their content, this is why they show up as different files with breakpoints.

One thing that could be done is to only show the breakpoints for the visible cells in the notebook, even though more exist as files on the backend. Some of this is discussed in https://github.com/jupyterlab/debugger/issues/171.

JohanMabille:
I think keeping the breakpoints for the whole notebook is pretty convenient. Another solution could be to "remember" when a cell gets the focus, and if its content is changed, first send a setBreakpoint request with an empty breakpoints list to the backend. This way the backend keeps the list of actual breakpoints, not the whole history.

KrzysztofSikoraCodete:
@JohanMabille Thanks for explanation

What we need is to get the path when the cell gets the focus, and remember it.
but only when the cell already has breakpoints.  We need kind of state pattern.

1. The cell get the focus, if breakpoints have been set, then enter for example "hold-on" state.
2. When no breakpoints then stay in "default state".
-  from "default state", everything works as now

From "hold-on" state:
[...]

PR #430, opened 2020-04-24
Extra typed breakpoints

KrzysztofSikoraCodete:
Suggestion to solve the problem

EDIT:
I tried to refer to the suggestion in #414 

Sending an empty list by setBreakpont is not a good solution. We lose references when we want to change the breakpoint in the cell.

Every path (id) of breakpoints are actually a reference to the story of all breakpoints in focused, active cell.
What I did is showing the last element from history of them.

My solution would showing breakpoints from active cell. Additionally i have to do a trigger for active cell. 

Focus event is not good idea, because doesn't work when user click on breakpoint gutter. However I'll check it.

Next, what i should do is clearing story after a certain number of queries, because all of them stay on memory.

Peek 2020-04-24 12-17

### DIFF
--- a/src/breakpoints/body.tsx
+++ b/src/breakpoints/body.tsx
@@ -65,13 +65,15 @@ const BreakpointsComponent = ({ model }: { model: BreakpointsModel }) => {
 
   return (
     <>
-      {breakpoints.map(entry => (
-        <BreakpointCellComponent
-          key={entry[0]}
-          breakpoints={entry[1]}
-          model={model}
-        />
-      ))}
+      {breakpoints
+        .filter(find => find[0] === model.currentBreakpointsId)
+        .map(entry => (
+          <BreakpointCellComponent
+            key={entry[0]}
+            breakpoints={entry[1]}
+            model={model}
+          />
+        ))}
     </>
   );
 };
--- a/src/breakpoints/model.ts
+++ b/src/breakpoints/model.ts
@@ -14,7 +14,7 @@ export class BreakpointsModel implements IDisposable {
   /**
    * Get the last breakpoints from story
    */
-  get currentBreakpoinstId(): string {
+  get currentBreakpointsId(): string {
     return this._currentBreakpointsId;
   }
   /**
--- a/src/breakpoints/model.ts
+++ b/src/breakpoints/model.ts
@@ -12,6 +12,18 @@ import { IDebugger } from '../tokens';
  */
 export class BreakpointsModel implements IDisposable {
   /**
+   * Get the last breakpoints from story
+   */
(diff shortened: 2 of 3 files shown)


jtpio:
Thanks @KrzysztofSikoraCodete.

Are you referring to #414?

KrzysztofSikoraCodete:
@jtpio  I've just edited description.

src/breakpoints/model.ts
```diff
@@ -11,6 +11,18 @@ import { IDebugger } from '../tokens';
  * A model for a list of breakpoints.
  */
 export class BreakpointsModel implements IDisposable {
+  /**
+   * Get the last breakpoints from story
```
jtpio, line 15: Does "story" here refer to the history of breakpoints?

KrzysztofSikoraCodete: Exactlly, `id` which is set here, i use to get current breakpoints cell's from story.

src/breakpoints/body.tsx
```diff
@@ -65,13 +65,15 @@ const BreakpointsComponent = ({ model }: { model: BreakpointsModel }) => {
...
     <>
-      {breakpoints.map(entry => (
-        <BreakpointCellComponent
-          key={entry[0]}
-          breakpoints={entry[1]}
-          model={model}
-        />
-      ))}
+      {breakpoints
+        .filter(find => find[0] === model.currentBreakpointsId)
```
jtpio, line 69: Does this mean that we display the breakpoints for the currently selected cell only?

KrzysztofSikoraCodete: Yes

jtpio: Or actually the last cell that was updated.

jtpio: I think we still want to show the breakpoints for the other cells. To be able to select the cell where the breakpoint is defined when they are clicked.

KrzysztofSikoraCodete: Peek 2020-04-24 12-58

How looks now.

KrzysztofSikoraCodete: Peek 2020-04-24 13-01

How works now

jtpio: When adding the breakpoint to the second cell the breakpoint list should show 3 breakpoints:

- 2 for the first cell
- 1 for the second cell

KrzysztofSikoraCodete: Sure, I'll figure something out

afshin:
Closed in favor of https://github.com/jupyterlab/debugger/pull/454

PR #436, opened 2020-05-05
Extra breakpoint after typing with refresh kernel.

KrzysztofSikoraCodete:
Peek 2020-05-05 14-57

My latest proposal to solve the problem. The solution is the implementation of the idea @JohanMabille.
https://github.com/jupyterlab/debugger/issues/414#issuecomment-621245446
Thanks for help.

What remains to be done.

I could delete the States class and send parameters directly, however they are correlated with cells. That's why I did it.

The second important thing that I haven't solved here is restoring points after refreshing the browser.

Naturally, the map is cleared, so it won't work in this case.

However, when refreshing the kernel, it seems to work.
The conclusion that comes to me. Some way to load breakpoints and create their map immediately after refreshing the browser.

I'm not sure about the final appearance updateBreakpoints method into service.

Any comments and suggestions?


src/breakpoints/model.ts
```diff
@@ -85,8 +91,38 @@ export class BreakpointsModel implements IDisposable {
   }
 
   private _isDisposed = false;
+  private _statesOfCell = new Map<string, string>();
```
jtpio, line 94: Do we need to keep track of extra state?

jtpio:
It looks like the diff shows changes from other merged PRs.

Should we copy the changes over to another new PR rebased on the latest maybe?

afshin:
Closing in favor of https://github.com/jupyterlab/debugger/pull/454

PR #438, opened 2020-05-07
Update the example notebook to include explanations and screenshots

jtpio:
Maybe it's a bit too much, but it sounds like it could be useful to have a more detailed example on how to use the debugger.

Something we could then add to the [jupyter-ide-demo](https://github.com/blink1073/jupyter-ide-demo) repo.

Screenshots are embedded as attachments in the notebook.

image


KrzysztofSikoraCodete:
Great idea. It will definitely be useful.

PR #443, opened 2020-05-15
Use signalToPromise from @jupyterlab/testutils in the tests

### LINKED ISSUE #306: Use signalToPromise from @jupyterlab/testutils
Instead of manually creating a `PromiseDelegate` to check for values emitted by `eventMessage`, we could try to use the `signalToPromise` helper function from `@jupyterlab/testutils`.

Here is an example:

https://github.com/jupyterlab/jupyterlab/blob/636a1aef5ff2908689b5550d64d9aed50e1d4bc8/tests/test-coreutils/src/restorablepool.spec.ts#L97

This could help clean up some patterns used in the tests, such as:

https://github.com/jupyterlab/debugger/blob/b95b79b8589690c28f78bf2b8dd64f59d6d709ff/tests/src/session.spec.ts#L238-L250

jtpio:
Fixes #306.

This helper function simplify the tests that wait for specific debug events from the kernel.


jtpio:
CI is green again after fixing the linting issue.

PR #428, opened 2020-04-21
Install libgbm1 to fix puppeteer in tests

afshin:
Install `libgbm1` to fix puppeteer.

PR #448, opened 2020-05-19
Remove the temporary libgbm1 fix

jtpio:
We should now be able to remove the `libgbm1` fix that was added in https://github.com/jupyterlab/debugger/pull/428 to fix the browser check.


jtpio:
CI is indeed passing now without the extra step.

PR #453, opened 2020-05-27
Add explicit return types to functions and methods

### LINKED ISSUE #451: Use stricter ESLint rules
The switch to ESLint was done in https://github.com/jupyterlab/debugger/pull/444.

The configuration files are mostly following the rules defined in core JupyterLab. However in this repo we can adopt stricter rules that would for example detect missing docstrings.

jtpio:
Fixes #451.

This will help catch missing docstrings and return types and future PRs.

This also helps refining what the debugger API should be.

### TODO

- [x] Explicit return types
- [x] JSDoc


src/service.ts
```diff
@@ -394,10 +385,20 @@ export class DebuggerService implements IDebugger, IDisposable {
     return reply.body.variables;
   }
 
+  /**
+   * Dump the content of a cell.
+   * @param code The source code to dump.
+   */
+  private async _dumpCell(
```
jtpio, line 392: This method is now private as it is only accessed from the service and is not part of the public `IDebugger` interface.

.eslintrc.js
```diff
@@ -47,7 +46,21 @@ module.exports = {
...
+          FunctionExpression: true,
+          ClassDeclaration: true,
+          MethodDefinition: true
+        }
+      }
+    ],
+    'jsdoc/require-param-type': 'off',
+    'jsdoc/require-property-type': 'off',
+    'jsdoc/require-returns-type': 'off',
+    'jsdoc/require-returns': 'off'
```
jtpio, line 63: Leaving these `off` for now.

jtpio: We can tackle them separately at some point if we want.

jtpio:
cc @afshin

jtpio:
cc @KrzysztofSikoraCodete we might want to get that in before rebasing the other PRs.

PR #441, opened 2020-05-13
Add an editor-finder plugin

KrzysztofSikoraCodete:
Refactor tracker.ts

Separated method find from tracker handler.
Created stand alone editor finder.

The plugin must be separated, instead of create the interface and token into tracker.ts, because if we don't do that will not be available in whole scope and will be a conflict with dependency cycle.

PR #454, opened 2020-05-28
Fix bug where additional breakpoints are added to a cell where a breakpoint has been triggered

KrzysztofSikoraCodete:
Solution prevents extra breakpoint after typing and more events.

https://github.com/jupyterlab/debugger/pull/436

### DIFF
--- a/src/editor-finder.ts
+++ b/src/editor-finder.ts
@@ -72,16 +72,18 @@ export class EditorFinder implements IDisposable, IDebuggerEditorFinder {
    *
    * @param debugSessionPath The path for the current debug session.
    * @param source The source to find.
+   * @param focus - Set to true to focus on the relevant cell. Default to false.
    */
   find(
     debugSessionPath: string,
-    source: string
+    source: string,
+    focus: boolean
   ): IIterator<CodeEditor.IEditor> {
     return chain(
-      this._findInNotebooks(debugSessionPath, source),
-      this._findInConsoles(debugSessionPath, source),
-      this._findInEditors(debugSessionPath, source),
-      this._findInReadOnlyEditors(debugSessionPath, source)
+      this._findInNotebooks(debugSessionPath, source, focus),
+      this._findInConsoles(debugSessionPath, source, focus),
+      this._findInEditors(debugSessionPath, source, focus),
+      this._findInReadOnlyEditors(debugSessionPath, source, focus)
     );
   }
 
@@ -90,10 +92,12 @@ export class EditorFinder implements IDisposable, IDebuggerEditorFinder {
    *
    * @param debugSessionPath The path for the current debug session.
    * @param source The source to find.
+   * @param focus - Set to true to focus on the relevant cell. Default to false.
    */
   private _findInNotebooks(
     debugSessionPath: string,
-    source: string
+    source: string,
+    focus: boolean
   ): CodeEditor.IEditor[] {
     if (!this._notebookTracker) {
       return [];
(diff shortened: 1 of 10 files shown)


jtpio:
> I am wondering why .find() is undefined in tests. Any suggestions? I think logic into service works correctly.

Maybe there could be a new test case that covers using the finder?

jtpio:
Testing this branch locally, it looks like the breakpoints are removed when switching between stack frames:

breakpoints-removed

src/handlers/notebook.ts
```diff
@@ -76,7 +77,8 @@ export class NotebookHandler implements IDisposable {
     const codeCell = cell as CodeCell;
     const editorHandler = new EditorHandler({
       debuggerService: this._debuggerService,
-      editor: codeCell.editor
+      editor: codeCell.editor,
+      idCell: cell.model.id
```
jtpio, line 81: We should now be able to remove references to the temporary cell ids and instead use the editor finder from https://github.com/jupyterlab/debugger/pull/441?

jtpio:
To avoid keeping track of extra state on the frontend, we probably want to implement something like this:

- the frontend gets the list of breakpoints from the kernel. This is part of the restore process but could also be retrieved manually by sending a `debugInfo` request
- the frontend gets the list of all breakpoints known by the kernel
- the frontend can check whether the breakpoints exist in any of the code editors currently visible (in a notebook, code console or file), by using the editor finder and the path (`source`) to the breakpoint
- if a path doesn't exist (for example it corresponds to a cell that has been removed or edited), then the path is deleted from the breakpoints map in the model.
- the breakpoints react component should pick up the change to the breakpoints model state and render the new list of breakpoints

KrzysztofSikoraCodete:
Peek 2020-06-08 15-28

Active cell bug solved.

src/editor-finder.ts
```diff
@@ -72,16 +72,18 @@ export class EditorFinder implements IDisposable, IDebuggerEditorFinder {
    *
    * @param debugSessionPath The path for the current debug session.
    * @param source The source to find.
+   * @param focus - Set to true to focus on the relevant cell. Default to false.
    */
   find(
     debugSessionPath: string,
-    source: string
+    source: string,
+    focus: boolean
```
jtpio, line 80: This looks like we could move the focus to a separate plugin, but it would be fine to keep it like this for now and refactor after.

src/service.ts
```diff
@@ -209,14 +213,96 @@ export class DebuggerService implements IDebugger, IDisposable {
...
+    };
+    return breakpointsForRestore(breakpoints, editorFinder);
+  }
+
+  /**
+   * Response to map of breakpoints
+   *
+   * @param reply - Response from restore state method.
+   */
+  async breakpointsMapFromServer(
```
jtpio, line 303: We probably want to choose a more descriptive name here (and make it private).

Also does this method need to set the hash parameters?

KrzysztofSikoraCodete: "Also does this method need to set the hash parameters?" im not sure. If we need to set the hash during calling updateBreakpoints method, then should be here. If not then we can leave this part of code only during restoreState method.

afshin: Hm, I don't see where else the hash parameters are set. This looks like the only spot. Does it belong here?

jtpio: Hashes should normally be set once per kernel during the restore process (retrieved from the debug info reply).

PR #455, opened 2020-06-03
Remove unused start and stop command ids

jtpio:
These are currently exported and available as `CommandIDs.start` and `CommandIDs.stop`, but not used for now.

Let's remove them and re-add back if / when needed.


jtpio:
Minor cleanup, merging.

Issue #220, opened 2019-11-22
Refreshing when stopped on a breakpoint


JohanMabille:
As pointed out by @jtpio in https://github.com/jupyterlab/debugger/pull/217, refreshing the page when the code execution is stopped on a breakpoint does not work. The reason is that after refreshing, the client sends a `kernel_info_request` to the kernel and waits for its reply. This request is sent on the shell, however the thread executing on the shell in `xeus_python` is stopped because of the previous breakpoint. Therefore, the client never gets its answer and loops on creating web sockets.

We can workaround this issue by restarting the debugger when refreshing the page. This would resume the execution of the thread on the `shell` without loosing the breakpoints.

A lont-term solution would be to have the `kernel_info_request` sent on the `control` instead of having it on the `shell` but that would require changes in Jupyter Lab core.

cc @afshin @KsavinN

jtpio:
It's also worth keeping an eye on the kernel and session architecture refactoring work happening in JupyterLab: https://github.com/jupyterlab/jupyterlab/pull/7252

For reference this is where the `kernel_info_request` message is sent on the `control` channel:

https://github.com/jupyterlab/jupyterlab/blob/00b17fc573af105044d03ca76d3f9cb07b871992/packages/services/src/kernel/default.ts#L478-L479

JohanMabille:
It seems that actaully the workaround is not possible since the web socket that allows to communicate with the kernel is not created until a first web socket receives a `kernel_info_reply`.

Therefore the only mean to fix that is to send the `kernel_info_request` on the control.

jtpio:
We discussed another way to detect whether a kernel supports debugging without having to rely on the debug info reply.

One workaround could be to send one of the debug requests, for example `debugInfo`, and wait for a response (without checking the content).

One possible issue with this is that other kernels such as `ipykernel` will receive `debug_request` messages and log a "Unknown message type" warning.

Also we currently use the `info` promise before sending messages to the kernel here:

https://github.com/jupyterlab/debugger/blob/dc94d28c6b27a8cbddd4bdf26a222e4d85e63c4d/src/session.ts#L197

Instead we could check whether the kernel is available with `this._connection?.kernel`;

Finally there is an issue about sending the `debug_info_request` on the control channel: https://github.com/jupyterlab/jupyterlab/issues/7560

jtpio:
Moving to 0.2.0 to check if this is a viable workaround.

jtpio:
[...]

PR #456, opened 2020-06-03
Handle page reload by sending a `debugInfo` to check whether a kernel supports debugging

jtpio:
Fixes https://github.com/jupyterlab/debugger/issues/220.

The issue is that at the moment the `kernel_info_request` message is sent on the `shell` channel. If the execution has stopped (breakpoint hit), the debugger frontend forever waits for the `kernel_info_reply` to decide whether the kernel supports debugging or not, because the `debugger` key is in the kernel info reply:

https://github.com/jupyterlab/debugger/blob/e41df5a0d57d2ec91fb91c04e1f41821a37e03d2/src/service.ts#L136-L139

Instead of waiting for the `kernel_info_reply`, the debugger frontend can send a `debug_request` message and check whether it gets a reply or not. `debugInfo` sounds like one of the best candidates for a `debug_request` message, as it can be sent at any time during a debug session.

This PR implements this idea:

refresh-on-stopped

This should also handle the case when the notebook is closed while the execution has stopped, and then re-opened again:

refresh-on-stopped-3

However, this brings the drawback of having the following message printed in the server logs when the kernel doesn't support debugging (for example ipykernel):

```
[IPKernelApp] ERROR | UNKNOWN CONTROL MESSAGE TYPE: 'debug_request'
```

This is because the debugger frontend now relies on sending at least one message to the kernel to check whether it handles debugging or not.

Another drawback is the fact that we need to set a timeout, as there is no error or reply when sending a message that a kernel doesn't recognize. As a side effect this introduces a little delay when switching kernels (here the timeout is set to 2s):

refresh-on-stopped-2

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -1,7 +1,9 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { Session } from '@jupyterlab/services';
+import { Session, KernelMessage } from '@jupyterlab/services';
+
+import { PromiseDelegate } from '@lumino/coreutils';
 
 import { IDisposable } from '@lumino/disposable';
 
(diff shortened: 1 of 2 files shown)


jtpio:
It might actually be better to specify whether a kernel supports debugging directly in the kernel spec? (this was also discussed / considered before)

It looks like the `metadata` field would allow arbitrary keys: https://jupyter-client.readthedocs.io/en/latest/kernels.html#kernel-specs

The debugger frontend would then use the `specsManager` to retrieve the kernel spec and decide to enable debugging or not. Something like the following:

```typescript
const name = 'xpython';
const debuggingEnabled = !!specsManager.specs!.kernelspecs[name]!.metadata?.['debugger'];
```

jtpio:
Replaced by https://github.com/jupyterlab/debugger/pull/457

PR #457, opened 2020-06-03
Use the kernel spec metadata['debugger'] to check if a kernel supports debugging

### LINKED ISSUE #220: Refreshing when stopped on a breakpoint

jtpio:
Fixes #220. 
Alternative to #456 (see this comment: https://github.com/jupyterlab/debugger/pull/456#issuecomment-638072880)

The issue is that at the moment the `kernel_info_request` message is sent on the `shell` channel. If the execution has stopped (breakpoint hit), the debugger frontend forever waits for the `kernel_info_reply` to decide whether the kernel supports debugging or not, because the `debugger` key is in the kernel info reply.

Instead of waiting for the `kernel_info_reply`, the debugger frontend now checks the kernel spec to know if a kernel supports debugging or not.

~~This change requires the following change to the kernel spec, and the kernels (for now only xeus-python) would have to implement that change too:~~

**EDIT**: this has been added to `xeus-python=0.8.0` (https://github.com/jupyter-xeus/xeus-python/pull/282):

```
cat ${CONDA_PREFIX}/share/jupyter/kernels/xpython/kernel.json
```

```json
{
  "display_name": "xpython",
  "argv": [
      "/home/jtp/miniconda/envs/debugger/bin/xpython",
      "-f",
      "{connection_file}"
  ],
  "language": "python",
  "metadata": { "debugger": true }
}
```

Visually the result looks similar as in #456, except that there is no need for a timeout anymore:

refresh-on-stopped-kernel-spec

refresh-on-stopped-5

This would also require an update to the sequence diagram in https://github.com/jupyterlab/debugger/issues/64.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { Session } from '@jupyterlab/services';
+import { Session, KernelSpec } from '@jupyterlab/services';
 
 import { IDisposable } from '@lumino/disposable';
 
@@ -25,14 +25,17 @@ import { VariablesModel } from './variables/model';
 export class DebuggerService implements IDebugger, IDisposable {
   /**
    * Instantiate a new DebuggerService.
+   *
+   * @param options The instantiation options for a DebuggerService.
    */
-  constructor() {
+  constructor(options: DebuggerService.IOptions) {
     // Avoids setting session with invalid client
     // session should be set only when a notebook or
     // a console get the focus.
     // TODO: also checks that the notebook or console
     // runs a kernel with debugging ability
     this._session = null;
+    this._specsManager = options.specsManager;
     this._model = new DebuggerModel();
   }
 
@@ -129,14 +132,15 @@ export class DebuggerService implements IDebugger, IDisposable {
    * @param connection The session connection.
    */
   async isAvailable(connection: Session.ISessionConnection): Promise<boolean> {
+    await this._specsManager.ready;
     const kernel = connection?.kernel;
     if (!kernel) {
       return false;
     }
-    const info =
-      (((await kernel.info) as unknown) as IDebugger.ISession.IInfoReply) ??
-      null;
-    return !!(info?.debugger ?? false);
+    const name = kernel.name;
+    return !!(
+      this._specsManager.specs.kernelspecs[name].metadata?.['debugger'] ?? false
+    );
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/service.ts
```diff
@@ -623,11 +627,28 @@ export class DebuggerService implements IDebugger, IDisposable {
...
+ */
+export namespace DebuggerService {
+  /**
+   * Instantiation options for a `DebuggerService`.
+   */
+  export interface IOptions {
+    /**
+     * The kernel specs manager.
+     */
+    specsManager: KernelSpec.IManager;
```
jtpio, line 648: We might want to make this optional, so instantiating a `DebuggerService` without a kernel specs manager would always assume debugging is enabled (for example in an cutom JupyterFrontEnd where only one kernel is available).

jtpio:
cc @JohanMabille @afshin 

Leaving it as a draft for now as it would require changes to `xeus-python`.

This is something we already discussed before. It looks like having the `debugger` key in the kernel spec metadata is cleaner than what is implemented in #456.

Also if we go for this or #456, the debugger key wouldn't be strictly needed in the `kernel_info_reply` anymore (https://github.com/jupyter/jupyter_client/pull/486).

afshin:
I think it's reasonable for a kernel that supports DAP to have metadata that says it supports debugging. This seems like the fastest route to get that information and it also seems the least kernel-traffic-dependent way of doing it. I support this approach over the other PR.

JohanMabille:
I totally agree with @afshin. I will implement the change in xeus-python quickly.

jtpio:
It looks like we need to mock the `KernelSpecManager` instance used in the tests, similar to what is done in core lab: https://github.com/jupyterlab/jupyterlab/blob/b1e2b83047421bf7196bec5f2a94d0616dcb2329/packages/services/test/kernelspec/manager.spec.ts#L17-L25

Retrieving the specs from the local jupyter server (child process) would otherwise return a 403, as the token is generated when the `JupyterServer` is started [here](https://github.com/jupyterlab/jupyterlab/blob/b1e2b83047421bf7196bec5f2a94d0616dcb2329/testutils/src/start_jupyter_server.ts#L205).

afshin (approved):
Thanks! This is a much nicer way to detect debugger functionality.

👍

kycutler:
Is this documented anywhere? I see the `debugger` flag documented in kernel info [here](https://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-info), but nothing about the flag in kernelspecs.

PR #459, opened 2020-06-03
Add tests for the debugger sidebar

### LINKED ISSUE #55: Add tests for the debugger UI
Once the interfaces and the overall structure become more stable, it will be useful to start adding tests for the UI elements as well.

This will help us test more complex scenarios and edge cases.

jtpio:
Partial fix for #55. 

Add a couple of tests to check that the breakpoints are rendered in the debugger sidebar.


jtpio:
We can add a couple of tests for the other sections too.

Issue #431, opened 2020-04-24
Add wheel tests back on CI


jtpio:
In https://github.com/jupyterlab/debugger/pull/415 we removed the GitHub Actions workflow that installs the `xeus-python` wheel from PyPI.

This is because of the way kernel specs are installed in the jest tests (tmp folders). See this comment for more info: https://github.com/jupyterlab/debugger/pull/415#issuecomment-619005997

We should however add the workflow back on CI.

jtpio:
Closing as the tests are now using the `xeus-python` wheel in core lab:

https://github.com/jupyterlab/jupyterlab/blob/fa86276692e559525a1e3e8f407e1d17b2adaf97/scripts/ci_install.sh#L53-L56

PR #462, opened 2020-06-04
Switch to JupyterServer from @jupyterlab/testutils

### LINKED ISSUE #445: Update the tests to use JupyterServer from @jupyterlab/testutils
The `JestApp` currently used in the tests here:

https://github.com/jupyterlab/debugger/blob/dfd6bf9d51a7a0cd78ca54f4173fcf527bd4d7fd/tests/run-test.py#L12

is now deprecated: 

https://github.com/jupyterlab/jupyterlab/blob/6d5240dda71fb3d12814e8b88b09e94e23d42875/jupyterlab/tests/test_app.py#L214

Instead we should switch to the new `JupyterServer` from `@jupyterlab/testutils`:

https://github.com/jupyterlab/jupyterlab/blob/6d5240dda71fb3d12814e8b88b09e94e23d42875/testutils/src/start_jupyter_server.ts#L34

jtpio:
Fixes #445. 

### Changes

- [x] Switch to `JupyterServer` from `@jupyterlab/testutils`, so there is no need for the `run-test.py` anymore and the deprecated `JestApp` (and tests become pure jest tests) 
- [x] Match the folder and test structure in core lab more closely


jtpio:
The jest reporter now looks the same as in core lab too (since it's reusing the same settings):

jest-reporting

jtpio:
It looks like kernel specs (in particular the `xpython` spec) don't have to be installed manually anymore, most likely because the environment patching doesn't override the existing jupyter paths.

jtpio:
For the record, this is the output of running `jupyter --paths` (before running the tests) on the GitHub Actions runner from the conda workflow:

```
config:
    /home/runner/.jupyter
    /usr/share/miniconda/envs/jupyterlab-debugger/etc/jupyter
    /usr/local/etc/jupyter
    /etc/jupyter
data:
    /home/runner/.local/share/jupyter
    /usr/share/miniconda/envs/jupyterlab-debugger/share/jupyter
    /usr/local/share/jupyter
    /usr/share/jupyter
runtime:
    /home/runner/.local/share/jupyter/runtime
```

jtpio:
We should be able to add the wheel test back on CI after this: https://github.com/jupyterlab/debugger/issues/431

PR #464, opened 2020-06-08
Unpin xeus

### LINKED ISSUE #452: Unpin xeus=0.23.14
`xeus` was pinned to `0.23.14` in https://github.com/jupyterlab/debugger/pull/441 so the tests could run locally and on CI.

See this specific comment for more info: https://github.com/jupyterlab/debugger/pull/441#issuecomment-634550512

We should check whether this pin is still needed after fixing https://github.com/jupyterlab/debugger/issues/445.

jtpio:
Fixes #452 

Follow-up to #463 as the pin shouldn't be needed anymore.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ jobs:
         conda init bash
         conda info -a
     - name: Create the conda environment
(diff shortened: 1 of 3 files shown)


jtpio:
Tests pass without the pin :heavy_check_mark: 

And `xeus=0.24.0` is installed as a dependency of `xeus-python=0.8.0`: https://github.com/jupyterlab/debugger/pull/464/checks?check_run_id=749704880

Issue #471, opened 2020-06-16
Extra breakpoints in code consoles and files


jtpio:
It looks like https://github.com/jupyterlab/debugger/pull/454 only fixes the case of cells in notebooks.

The behavior described in #414 is still reproducible with code consoles and files:

![extra-breakpoints-console](https://user-images.githubusercontent.com/591645/84752550-2d7c7680-afbe-11ea-8d71-41bceaf20b89.gif)

![extra-breakpoints-file](https://user-images.githubusercontent.com/591645/84752567-30776700-afbe-11ea-8c74-2f093ae42ae6.gif)

Tested on Binder using master (with `0.3.0-beta.1`):

https://mybinder.org/v2/gh/jupyterlab/debugger/881ed8d9368e4d9cb700bde1064407095bc562a4?urlpath=/lab/tree/examples/index.ipynb

jtpio:
Fixed by #474.

Issue #467, opened 2020-06-13
Refactor DebuggerService#_filterBreakpoints method


afshin:
This method's logic is difficult to reason about and can be simplified.

KrzysztofSikoraCodete:
Fixed by #414

PR #474, opened 2020-06-17
Store configuration options for multiple kernels

KrzysztofSikoraCodete:
Related with issue:
https://github.com/jupyterlab/debugger/issues/471

Summary:

Added editor finder to console and file handler


KrzysztofSikoraCodete:
Peek 2020-06-18 12-15

KrzysztofSikoraCodete:
@jtpio @afshin 
I've separated out the hash methods.
Now the structure seems more sensible and tests are passing.

KrzysztofSikoraCodete:
I found some issue after initialization parameters-mixer plugin, although tests passed

KrzysztofSikoraCodete:
Solved

KrzysztofSikoraCodete:
Solved https://github.com/jupyterlab/debugger/issues/467 also on this PR

KrzysztofSikoraCodete:
> We should do a follow-on PR where the hash also is associated with a kernel name.

I've added kernel name as third param.

src/debugger-configuration.ts
```diff
@@ -0,0 +1,102 @@
...
+  public setTmpFileParameters(
+    prefix: string,
+    suffix: string,
+    kernelName: string
+  ): void {
+    this._tmpFileAssociatedWithKernel.set(kernelName, [prefix, suffix]);
+  }
+
+  private _hashMethod: (code: string) => string;
+  private _tmpFileAssociatedWithKernel = new Map<string, [string, string]>();
```
jtpio, line 85: Maybe there could be an interface instead of `[string, string]`?

So it's easier to follow what they refer to.

src/debugger-configuration.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Interface for configuration plugin
+ */
+export interface IDebuggerConfig {
+  setHashParameters(method: string, seed: number, kernelName: string): void;
+  setTmpFileParameters(
+    prefix: string,
+    suffix: string,
+    kernelName: string
+  ): void;
```
jtpio, line 100: We probably want to pass an object with the `method`, `seed` and `kernelName` properties instead of multiple arguments to the function, so it's easier to extend later if needed.

afshin:
I agree with the latter point. The former point also seems like a good
idea. I'll be pushing an update to this.

jtpio:
We should now be able to rebase the PR to fix the tests.

I'll give the PR a quick test locally too.

src/editor-finder.ts
```diff
@@ -93,11 +95,13 @@ export class EditorFinder implements IDisposable, IDebuggerEditorFinder {
...
    * @param source The source to find.
    * @param focus - Set to true to focus on the relevant cell. Default to false.
+   * @param kernelName The kernel name from current session.
    */
   private _findInNotebooks(
     debugSessionPath: string,
     source: string,
-    focus: boolean
+    focus: boolean,
+    kernelName: string
```
jtpio, line 104: Similar comment as above. These `findIn*` methods used to have 2 arguments only, but now that they are growing we might want to refactor that into an object too.

src/debugger-configuration.ts
```diff
@@ -0,0 +1,102 @@
...
+// Distributed under the terms of the Modified BSD License.
+
+import { Token } from '@lumino/coreutils';
+
+import { Signal } from '@lumino/signaling';
+
+import { murmur2 } from 'murmurhash-js';
+
+/**
+ * A class to hash code.
```
jtpio, line 11: How about something more specific?

```suggestion
 * A class that holds debugger configuration for a kernel.
```

jtpio:
Just tried locally and it looks good, thanks @KrzysztofSikoraCodete :+1: 

Having the debug config (hash functions and parameters) specific to a kernel indeed makes sense.

> I'll be pushing an update to this.

@afshin looks like we should be able to merge this PR in once rebased and updated, so it can be included in the next release.

src/debugger-configuration.ts
KrzysztofSikoraCodete: It makes sense

KrzysztofSikoraCodete:
Nice to hear that. I'll open new PR after merge this one, then i'll improve about what in the above remarks.

jtpio:
> I'll open new PR after merge this one, then i'll improve about what in the above remarks

I thought we could do it in the same PR here.

KrzysztofSikoraCodete:
> I thought we could do it in the same PR here.

Right, so i'll do here

KrzysztofSikoraCodete:
Excellent idea. Thanks @afshin for refactor and cleaning

PR #481, opened 2020-07-02
Remove Font Awesome CSS

### LINKED ISSUE #480: Fontawesome 4 css
Having this css that is very specific to fontawesome 4 is messing things up in jlab 2, where we've upgraded to fontawesome 5:

https://github.com/jupyterlab/debugger/blob/38d724014319aa54cd287d1d231632e188a89849/style/index.css#L15-L21

It overrides the global `.fa` class so that the fontawesome 5 shim does not properly work, and I think it breaks fontawesome for everyone else that is using the `.fa` class in jlab 2.

jtpio:
Fixes #480. 

Looking at the git history it seems like this is a leftover from the very early iterations that didn't get cleaned up.

### DIFF
--- a/style/index.css
+++ b/style/index.css
@@ -10,20 +10,6 @@
 @import './sidebar.css';
 @import './sources.css';
 
-/* font awesome */
-
-.fa {
-  display: inline-block;
-  font: normal normal normal 14px/1 FontAwesome;
-  font-size: inherit;
-  text-rendering: auto;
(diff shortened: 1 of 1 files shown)


jtpio:
cc @afshin it would be nice if we could release a new version with this change, as it has side effects on other extensions (https://github.com/jupyterlab/debugger/issues/480#issue-649478121)

PR #484, opened 2020-07-08
Fix tests

### LINKED ISSUE #485: Add yarn.lock
In the very early iterations it was [discussed](https://github.com/jupyterlab/debugger/pull/6) whether or not check the `yarn.lock` file in (it makes sense to not add during the early stages)

We should now be able to add it to the repo. This would also help debug issues like this one: https://github.com/jupyterlab/debugger/pull/484

jtpio:
This is an attempt at fixing the tests, which are not compiling anymore due to a conflict in the type definitions of `@jupyterlab/services` (multiple versions).

See https://github.com/jupyterlab/jupyterlab/issues/8628 for more details.

**Edit**: fixes #485 too


jtpio:
This looks new:

```
TypeError: request to http://localhost/api/sessions?1594199926031 failed, reason: connect ECONNREFUSED 127.0.0.1:80FetchError: request to http://localhost/api/sessions?1594199926031 failed, reason: connect ECONNREFUSED 127.0.0.1:80
```

jtpio:
Similar behavior locally, trying to connect to `http://localhost/api` instead of `http://localhost:8888/api`.

jtpio:
It looks like the default server settings are not able to fetch the correct `baseUrl` that should be set by `JupyterServer` here: 

https://github.com/jupyterlab/jupyterlab/blob/abc0dd4255a037af37828122bd2e199857fd64db/testutils/src/start_jupyter_server.ts#L269

jtpio:
Updating the dependencies so they all require the `-rc.1` versions seems to fix the issue.

jtpio:
It seems to be related to the version of `@jupyterlab/coreutils` (which exposes `PageConfig`).

`JupyterServer` would set the server `baseUrl` with `PageConfig.setOption('baseUrl', baseUrl)`, but the tests would not be able to pick it up with `PageConfig.getBaseUrl()` when creating the server connection settings here:

https://github.com/jupyterlab/jupyterlab/blob/47fc92a8d35b5c1346f9ea1390fbd27fbf3a38ea/packages/services/src/serverconnection.ts#L218

2 commits to show the difference:

- it fails with `"@jupyterlab/coreutils": ^4.0.0"`: https://github.com/jupyterlab/debugger/pull/484/checks?check_run_id=849788850
- it passes with `"@jupyterlab/coreutils": ^4.2.0-rc.1"`: https://github.com/jupyterlab/debugger/pull/484/checks?check_run_id=849805740

PR #486, opened 2020-07-08
Fix file editor toolbar CSS

### LINKED ISSUE #477: Make sure we aren't clobbering relevant toolbar styling
cf. https://github.com/elyra-ai/elyra/issues/671

jtpio:
Fixes #477 

This change removes the styling applied to the document toolbar.

Since the toolbar is by default empty (although third-party extensions could add items), we might as well keep it at the start, the same for the console.

### Before

image

### After

image

### DIFF
--- a/style/console.css
+++ b/style/console.css
@@ -1,7 +1,0 @@
-.jp-ConsolePanel::before {
-  content: none;
-}
-
-.jp-ConsolePanel .jp-Toolbar {
-  justify-content: flex-end;
-}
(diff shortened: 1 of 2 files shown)


jtpio:
Also had a quick look at the rest of the CSS, and the remaining rules look good now and shouldn't have side-effects on the rest of the lab application.

The `style/console.css` file was setting general rules:

https://github.com/jupyterlab/debugger/blob/c69e05c6a2c71d421d67e9af60b9ec2d206d6ee1/style/console.css#L1-L7

But this PR removes it.

PR #497, opened 2020-07-15
Update to JupyterLab 2.2 final

jtpio:
JupyterLab 2.2 was released today, so we can update the dependencies to `^2.2.0`.


jtpio:
Merging as mechanical change.

PR #501, opened 2020-07-21
Update interfaces and token definitions

### LINKED ISSUE #450: Fix casting of IDebugger.IModel as DebuggerModel in debugger.ts
`this.model = service.model as DebuggerModel;` should not be necessary in `debugger.ts`

### LINKED ISSUE #495: Read Only EditorTracker is instantiated twice
As a side-effect of the refactoring of the editor finder, and the split from the tracker plugin, it looks like the `readOnlyEditorTracker` is now instantiated twice:

https://github.com/jupyterlab/debugger/blob/f446f56062528a2523a343f17aae8ca5607108fd/src/handlers/tracker.ts#L57-L61

https://github.com/jupyterlab/debugger/blob/f446f56062528a2523a343f17aae8ca5607108fd/src/editor-finder.ts#L39-L42

This means that the tracker plugin is not able to find existing read only editors:

read-only-editor-tracker

jtpio:
Fixes #495.
Fixes #450. 

This PR makes the following changes:

- `EditorFinder` now owns the read-only editor tracker. This seems the most natural option for now since the `EditorFinder` needs access to it to find the `CodeEditor.IEditor` in all the tracked widgets.
- Plugins that want to open read-only views of the sources must call `IEditorFinder#open`. The editor finder opens the editor wrapper in the main area and adds the `MainAreaWidget` to its internal tracker (so it can be looked up with the `find` method).
- For end-users, this restores the following functionality:

fix-read-only-tracker

One thing to notice with this approach is the fact that the read-only editor tracker (`WidgetTracker<MainAreaWidget<CodeEditorWrapper>>`) stays private to the `IEditorFinder` plugin. Later we might want to expose the tracker so other plugins can use and extend it.


jtpio:
If `EditorFinder` manages the sources (`find`, `open` and focus), we might also want to use a more explicit name (potentially with "sources" in it).

---

We should also consider renaming the `tracker` plugin:

https://github.com/jupyterlab/debugger/blob/4f56bb73da45e758416bd7a2c743f901cebfef72/src/index.ts#L250

And `TrackerHandler`: 

https://github.com/jupyterlab/debugger/blob/4f56bb73da45e758416bd7a2c743f901cebfef72/src/index.ts#L259

The scope of the `tracker` plugin now is to use the editor finder to find sources or open them in read-only editors when:

- a breakpoint is clicked
- the "Open in main area" source button is clicked

(but this can be done separately)

src/handlers/tracker.ts
```diff
@@ -219,27 +191,5 @@ export namespace TrackerHandler {
...
-   */
-  export type IDebuggerReadOnlyEditorTracker = IWidgetTracker<
-    MainAreaWidget<CodeEditorWrapper>
-  >;
-
-  /**
-   * The Debugger Read Only Editor tracker token.
-   * TODO: provide the token for the tracker in the plugin?
-   */
-  export const IDebuggerReadOnlyEditorTracker = new Token<
```
jtpio, line 242: This token is not necessary anymore if the tracker stays private.

afshin:
Should this still be called `IEditorFinder`? What does it do?

* It keeps track of all the editors in all of the open documents that support debugging.
* It allows you to `find` a relevant editor.
* It allows you to `open` a read-only editor in the UI.

I'm not thinking of a clever name off the bat, but now is one of our last chances to name this well before we have API lock-in. What do you think?

jtpio:
The name should indeed change.

> If `EditorFinder` manages the sources (`find`, `open` and focus), we might also want to use a more explicit name (potentially with "sources" in it).

It looks like another option could include `Sources` in it, since this is about finding existing source cells, or opening them in read-only editors. But keeping `Editor` might make more sense for now since this corresponds to the API boundaries (retrieving editors or adding new ones).

JohanMabille:
EditorManager? (if in doubt, name it manager ;))

afshin:
`SourceSource`, it's the source for sources.

jtpio:
> EditorManager? (if in doubt, name it manager ;))

This is tempting :) But often manager is too "generic" and over time we forget what it manages.

afshin:
`DebuggerSources` and `IDebuggerSources` maybe?

`DebuggerSources#find()` and `DebuggerSources#open()` both make some sense.

jtpio:
Sounds like it would work yes :+1: 

Then the plugin can be renamed to `@jupyterlab/debugger:sources`

afshin:
And the file name could be `sources.ts`. Hm, I don't ❤️ this name but it's a closer match to the API and functionality.

jtpio:
But this file could be moved into another subfolder.

afshin:
Ah yes, you are correct. Hm.

jtpio:
We are also renaming a couple of other components in this PR (not only the editor finder), to provide a more consistent way of consuming them.

src/tokens.ts
```diff
@@ -142,13 +143,215 @@ export interface IDebugger {
...
 }
 
 /**
  * A namespace for visual debugger types.
  */
 export namespace IDebugger {
+  /**
+   * A namespace for UI interface definitions.
+   */
+  export namespace UI {
```
jtpio, line 156: Wondering whether this should be UI specific.

Could `Model` also be an adequate name?

afshin: Yeah, I think that makes sense, too.

jtpio:
Looks like the latest test failure is relevant (tests might need to be updated): https://github.com/jupyterlab/debugger/pull/501/checks?check_run_id=896507154

afshin:
I think we should remove the `TrackerHandler` and its plugin completely and move its logic into the `DebuggerService`. I haven't fully investigated, but this seems possible now.

jtpio:
The original logic that was part of the `TrackerHandler` has been progressively moved to different plugins, and it looks like it would make sense to do the same with the remaining bits.

jtpio:
Could the `sidebar` plugin be another candidate for this? Not sure the `DebuggerService` would be the best fit, if we aim at keeping it UI agnostic.

afshin:
That's a good point, yes.

jtpio:
It would also be fine to do that change in a separate PR.

PR #505, opened 2020-07-22
Hide the sidebar if no kernel supports debugging

### LINKED ISSUE #289: Toggle button and sidebar visibility

jtpio:
Fixes #289.

hide-sidebar

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -407,7 +407,18 @@ const main: JupyterFrontEndPlugin<void> = {
     settingRegistry: ISettingRegistry | null,
     themeManager: IThemeManager | null
   ): Promise<void> => {
-    const { commands, shell } = app;
+    const { commands, shell, serviceManager } = app;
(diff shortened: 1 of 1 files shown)


jtpio:
This approach of just hiding the sidebar (or rather not adding it) instead of disabling the plugin all together (for example by not providing the `IDebugger` token), means that there is still a hint about the debugger extension in the settings:

image

Issue #493, opened 2020-07-14
Prepare public interfaces for merging into JupyterLab core


afshin:
- [ ] **MAYBE** `ISession` should not expose any of its start/stop/etc. functionality and instead the `IDebugger` interface should pass those methods through. We should discuss this further.
- [x] `IEditorFinder#find` should return an array instead of an iterator

jtpio:
Also related and maybe worth considering: https://github.com/jupyterlab/debugger/issues/450

jtpio:
Looking into https://github.com/jupyterlab/debugger/issues/495, it looks like we could also removed `focus` from `IEditorFinder.Params`, so the editor finder doesn't have any side-effect on the widgets, and is only responsible for finding the code editors.

This would require the caller of `IEditorFinder#find` to implement the focus logic, for example:

https://github.com/jupyterlab/debugger/blob/4f56bb73da45e758416bd7a2c743f901cebfef72/src/editor-finder.ts#L94-L100

But we would need a way to find the notebook, console or file editor widget from a `CodeEditor.IEditor`.

jtpio:
It should now also be possible to remove the `tracker` plugin:

https://github.com/jupyterlab/debugger/blob/9698d6f252ce4a366b7300a57e32cfbe0326b92a/src/index.ts#L233-L234

And move the remaining of the logic: 

https://github.com/jupyterlab/debugger/blob/9698d6f252ce4a366b7300a57e32cfbe0326b92a/src/handlers/tracker.ts#L45-L63

To the `main` plugin:

https://github.com/jupyterlab/debugger/blob/9698d6f252ce4a366b7300a57e32cfbe0326b92a/src/index.ts#L390

See discussions in https://github.com/jupyterlab/debugger/pull/501 for more context.

jtpio:
- [ ] Also, this alias could be either removed or moved to the token definitions:

https://github.com/jupyterlab/debugger/blob/9698d6f252ce4a366b7300a57e32cfbe0326b92a/src/panels/callstack/model.ts#L79

jtpio:
> Also, this alias could be either removed or moved to the token definitions:

It could probably be moved to `tokens.ts`, similar to `IDebugger.IBreakpoint`:

https://github.com/jupyterlab/debugger/blob/20efafc9eac4ded21a14d5a6e668f205e1fe056e/src/tokens.ts#L176

And named `IStackFrame`.

Or as a type alias:

```typescript
[...]

PR #507, opened 2020-07-23
Remove the tracker plugin

jtpio:
Addressed the following comment: https://github.com/jupyterlab/debugger/issues/493#issuecomment-662931028

By removing the `tracker` plugin and moving its logic to the `main` plugin.


afshin:
Should `IDebuggerSources` be `optional` here? Is there a time when it won't be available?

jtpio:
We could imagine a lab build that disables `@jupyterlab/debugger:sources` for example (to disable the opening of read-only editors in the main area).

Also it's general it looks like it can be more flexible to stick to `optional` by default to avoid hard requirements?

Although for now `IDebuggerSources` is still a `require` for `IDebugger`:

https://github.com/jupyterlab/debugger/blob/9698d6f252ce4a366b7300a57e32cfbe0326b92a/src/index.ts#L258

But could be made `optional` too since its options do so:

https://github.com/jupyterlab/debugger/blob/9698d6f252ce4a366b7300a57e32cfbe0326b92a/src/service.ts#L668

afshin:
Yeah that makes sense.

afshin:
Do you want to make it `optional` for `IDebugger` in this PR? Or should we merge this one?

jtpio:
Yes I'll push that now

PR #509, opened 2020-07-23
Move CallstackModel.IFrame to IDebugger.IStackFrame

jtpio:
Addresses this comment: https://github.com/jupyterlab/debugger/issues/493#issuecomment-662955582

### DIFF
--- a/src/tokens.ts
+++ b/src/tokens.ts
@@ -167,7 +167,7 @@ export namespace IDebugger {
   /**
    * The type for a stack frame
    */
-  export type IStackFrame = DebugProtocol.StackFrame;
+  export interface IStackFrame extends DebugProtocol.StackFrame {}
 
   /**
    * Single breakpoint in an editor.
--- a/src/tokens.ts
+++ b/src/tokens.ts
@@ -13,10 +13,6 @@ import { ISignal, Signal } from '@lumino/signaling';
 
 import { DebugProtocol } from 'vscode-debugprotocol';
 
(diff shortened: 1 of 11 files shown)


jtpio:
We might want to do the same with `VariablesModel`:

https://github.com/jupyterlab/debugger/blob/20efafc9eac4ded21a14d5a6e668f205e1fe056e/src/tokens.ts#L681-L686

src/tokens.ts
```diff
@@ -170,13 +164,43 @@ export namespace IDebugger {
     path: string;
   };
 
+  /**
+   * The type for a stack frame
+   */
+  export type IStackFrame = DebugProtocol.StackFrame;
```
afshin, line 170: Maybe we should do `interface IStackFrame extends DebugProtocol.StackFrame{}` if we want the `I` at the beginning? I'm not sure what to do here, to be honest.

jtpio: Yes eslint formats it automatically. But we could disable the rule.

afshin: Oh right, I forgot that constraint ... from yesterday :joy:

Hm, I think we should disable that rule if it's making programmatic decisions for us. What do you think?

jtpio: Sounds good :+1: 

Especially since core lab disables it too: https://github.com/jupyterlab/jupyterlab/blob/5e34867f355f63a5dcf5f77e6ff780f877b5748f/.eslintrc.js#L38

PR #511, opened 2020-07-23
Throw if the config does not have tmp file params

jtpio:
Similar to the `hash` function look-up in the `DebuggerConfig`.

Since the `DebuggerConfig` relies on the caller to use `setTmpFileParams` and `setHashParams`, it can also throw an error if the tmp file parameters are not set.

This change also adds basic unit tests for the `DebuggerConfig`.

### DIFF
--- a/test/config.spec.ts
+++ b/test/config.spec.ts
@@ -9,6 +9,15 @@ describe('DebuggerConfig', () => {
   });
 
   describe('#getCodeId', () => {
+    it('should compute a valid code id when the parameters are set', () => {
+      const [prefix, suffix] = ['foo', 'bar'];
+      config.setHashParams({ method: 'Murmur2', seed: 'bar', kernel });
+      config.setTmpFileParams({ prefix, suffix, kernel });
+      const codeId = config.getCodeId('i = 0', kernel);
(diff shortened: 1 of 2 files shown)


test/config.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+      expect(() => {
+        config.getCodeId('i = 0', kernel);
+      }).toThrow('has no hashing params');
+    });
+
+    it('should throw if the kernel does not have tmp file parameters', () => {
+      config.setHashParams({ method: 'Murmur2', seed: 'bar', kernel });
+      expect(() => {
+        config.getCodeId('i = 0', kernel);
+      }).toThrow('has no tmp file params');
```
afshin, line 23: Does `toThrow(str)` do an `indexOf` to match that string? Because this string is a substring of the actual thrown error. I'm curious what it is passing.

jtpio: Yes, there are some examples here: https://jestjs.io/docs/en/expect#tothrowerror

But we might actually want to pass the full message to be more explicit.

afshin: That makes sense. It is okay to leave it as is, in that case, I think.

PR #513, opened 2020-07-24
Improve the handling of closed notebooks

### LINKED ISSUE #510: TypeError: can't access property "name", service.session.connection.kernel is null
This error is logged in the dev tools console, which seems to be happening during the book-keeping of the breakpoints:

console-error-breakpoints

jtpio:
Fixes #510.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -124,7 +124,7 @@ export class DebuggerService implements IDebugger, IDisposable {
    * @param code The source code.
    */
   getCodeId(code: string): string {
-    return this._config.getCodeId(code, this.session.connection.kernel.name);
+    return this._config.getCodeId(code, this.session?.connection?.kernel?.name);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/service.ts
```diff
@@ -124,7 +124,14 @@ export class DebuggerService implements IDebugger, IDisposable {
...
   getCodeId(code: string): string {
-    return this._config.getCodeId(code, this.session.connection.kernel.name);
+    try {
+      return this._config.getCodeId(
+        code,
+        this.session?.connection?.kernel?.name
+      );
+    } catch {
+      return '';
+    }
```
jtpio, line 134: Not the most elegant, but this ensures the return type of `DebuggerService#getCodeId` stays a `string`.

jtpio:
The last commit also handles the case of non-existing kernels (for example when switching to non-existing kernels), where the model wouldn't be cleared and the sidebar still populated:

switch-kernel-null

PR #520, opened 2020-07-27
Return true from DebuggerServicer#isAvailable when spec is not available

mnowacki-b:
This is a followup to: https://github.com/jupyterlab/debugger/pull/518

Returning `true` when calling `DebuggerServicer#isAvailable` and the `this._specsManager.specs.kernelspecs` are not available for the current kernel.


jtpio:
Thanks @mnowacki-b!

jtpio:
Also related to https://github.com/jupyterlab/jupyterlab/pull/8727

Issue #526, opened 2020-07-31
Add default configuration override to show extension by default


mnowacki-b:
Currently if there is not kernel spec available, that have metadata with 'debugger' flag set, the extension will be disabled (see https://github.com/jupyterlab/debugger/blob/master/src/index.ts#L387 ). 

It would be nice to have an override where we could enable/disable the extension by default even if the specs are not available.

jtpio:
Fixed by https://github.com/jupyterlab/debugger/pull/527

PR #527, opened 2020-07-31
Add page config override for debugger panel

mnowacki-b:
Related to https://github.com/jupyterlab/debugger/issues/526

Adding PageConfig override to enable / disable the debugger extension.

If the override is set to true: enable the extension without checking kernel specs.
If the override is set to false: disable the extension.
If the override is not set: check the specs to determine whether or not the extension should be enabled.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -397,13 +397,10 @@ const main: JupyterFrontEndPlugin<void> = {
     const { kernelspecs } = serviceManager;
 
     // First check if there is a PageConfig override for the extension visibility
-    const showDebuggerExtension = String(
-      PageConfig.getOption('showDebuggerExtension')
-    ).toLowerCase();
-    if (showDebuggerExtension === 'false') {
-      return;
-    }
-    if (showDebuggerExtension !== 'true') {
+    const alwaysShowDebuggerExtension =
+      PageConfig.getOption('alwaysShowDebuggerExtension').toLowerCase() ===
+      'true';
+    if (!alwaysShowDebuggerExtension) {
       // hide the debugger sidebar if no kernel with support for debugging is available
       await kernelspecs.ready;
       const specs = kernelspecs.specs.kernelspecs;
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,7 +19,7 @@ import { IEditorServices } from '@jupyterlab/codeeditor';
 
 import { ConsolePanel, IConsoleTracker } from '@jupyterlab/console';
 
-import { PathExt } from '@jupyterlab/coreutils';
+import { PageConfig, PathExt } from '@jupyterlab/coreutils';
 
 import { DocumentWidget } from '@jupyterlab/docregistry';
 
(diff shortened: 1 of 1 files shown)


mnowacki-b:
@afshin @jtpio 
Please take a look
Thank you!

SylvainCorlay:
Thanks!

afshin:
Hi @mnowacki-b! I think this functionality makes sense as a settable behavior of the debugger.

I think perhaps the way we should do this is to have a default setting for kernel auto-detection that is set to `true` but can be turned off with `overrides.json` or a user preference instead of passing this information via page config, though.

What do you think?

afshin:
Also, please note that we have merged the debugger into JupyterLab, so we could add this feature for the JupyterLab 3.0 release as well. Or we could do it in both versions.

mnowacki-b:
Hi @afshin 
Thank you for your suggestion!
I am not too familiar with `overrides.json` can you point me to documentation and/or source code so I can take a look at it?

Yes we were planing on porting it over into the JupyterLab once everything is ironed out.

mnowacki-b:
OK I found the info on `overrides.json` it is used by the `SettingRegistry` that was suggested by @jtpio as well.
Maybe that will be a better way to go.

mnowacki-b:
Isn't `SettingRegistry` more like a dynamic configuration where a user can change those setting during runtime? Where as PageConfig is more of a startup configuration which kinds of fits for this case?
Do we want to have ability to turn on/off the debugger visibility during runtime?

afshin:
The setting registry is indeed for dynamic configurations that users can change. It has three tiers, though:

0. `overrides.json` allows a deployer/administrator to decide what the **default** value of a preference is.
1. The default values in the settings schema allows an extension author to set a user preference's default.
2. The end user can set whatever value they wish.

I'm a little concerned about having multiple ways to detect whether the debugger extension gets rendered on the screen or not and I wanted the end user to have a way out of a potentially incorrect situation.

I think the real solution is for the kernel in question to add the relevant metadata to its kernelspec, though. So basically I'm a little ambivalent about the correct way to deal with a kernel that supports Jupyter's version of DAP but does not say so in the kernelspec.

@SylvainCorlay @jasongrout, do you have an opinion about what to do here?

mnowacki-b:
@afshin 
I guess we have a unique case where debugger kernel specs are not available in the spec manager; however, we are still able to connect to a debugger kernel using a customs extension. That is why it would be nice to be able to enable the extension by default without any further checks.
I can make the changes to the PR to use setting registry if that would be a more suitable path forward?

SylvainCorlay:
I think that the point of @mnowacki-b's PR is to have an option to *always* show the extension even if none of the listed kernels supports debugging.

SylvainCorlay:
It can probably be achieved with overrides, like `always_show_debugger_extension`...

mnowacki-b:
@afshin
If we go with the overrides do you think that would be acceptable?

afshin:
I think you're right that page config is a sensible place for this. I'm still thinking about the name of the flag, something like `alwaysShowDebuggerExtension` like Sylvain suggests seems clearer.

The only other comment is that you don't need the `String(...)` constructor because `getOption` always returns a string: 
```typescript
const alwaysShow = PageConfig.getOption('alwaysShowDebuggerExtension').toLowerCase() === 'true';
```

Let's go with your approach.

jasongrout:
> @SylvainCorlay @jasongrout, do you have an opinion about what to do here?

My initial thought is that the existing logic is sufficient. What are the situations where we want the debugger to be shown even if no kernel spec claims support for debugging?

> I guess we have a unique case where debugger kernel specs are not available in the spec manager

It feels like the right way to deal with this specific case is to put the debugger kernel spec in the kernelspec manager, where the debugger extension will see it and activate itself. I think we have a lot of assumptions throughout the code that the kernel spec manager has all the kernel specs available on the server.

I think the awkwardness here comes from trying to solve the confusing problem of having two python kernels which are not quite equivalent, and only one of which supports debugging. Is there a way to hide (or at least make automatic) a kernel change when debugging is turned on? Could there be a way to nominally have one kernel (the ipython kernel), but automatically launch the xeus kernel when debugging is turned on? This would be similar to having a debug configuration you have to run in a more traditional IDE, which spins up a different process (or at least different flags) from the standard process. In this case, though, the debugger would not only see if the current kernel supported debugging, but would also see if an equivalent kernel supported debugging, and automatically switch to it when debugging was turned on.

SylvainCorlay:
> Let's go with your approach.

:+1: I think it is fine to allow a setting to always show the debugger extension.

mnowacki-b:
@afshin @jasongrout @saulshanabrook 
I have update the code as suggested
- rename the config to `alwaysShowDebuggerExtension`
- small refactor in how we use the setting
   * if the setting is set to 'true' - bypass the kernel spec check and show the extension
   * else - show the extension based on original behavior (check the kernel spec)

afshin:
Let's merge this and iterate of @jasongrout has a comment about how to do it differently. If not, I'll publish a patch release tomorrow (Wednesday) that includes this.

mnowacki-b:
Thank you!

PR #532, opened 2020-08-19
Condition for empty string

KrzysztofSikoraCodete:
Solution for problem with setting and removing breakpoints. Condition has been changed for case with empty string of `the_path`


afshin (approved):
Cheers!

PR #533, opened 2020-08-27
Adding an optional custom panel to the sidebar

mnowacki-b:
Adding ability to extend the Sidebar with a custom panel that will be displayed at the top (see screenshot below).

Screen Shot 2020-08-26 at 7 17 29 PM

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -46,11 +46,18 @@ import { DebuggerHandler } from './handler';
 
 import { EditorHandler } from './handlers/editor';
 
-import { IDebugger, IDebuggerConfig, IDebuggerSources } from './tokens';
+import {
+  IDebugger,
+  IDebuggerConfig,
+  IDebuggerSources,
+  IDebuggerSidebar
+} from './tokens';
 
 import { ReadOnlyEditorFactory } from './panels/sources/factory';
 
 import { VariablesBodyGrid } from './panels/variables/grid';
+
+export { IDebugger, IDebuggerSidebar } from './tokens';
 
 /**
  * The command IDs used by the debugger plugin.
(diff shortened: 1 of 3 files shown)


afshin:
Hello @mnowacki-b! We should not add any more features to the JupyteLab 2.x `@jupyterlab/debugger` 0.3.x line and we should add them directly to JupyterLab core since we are about to do release that version. This repository is on its way to being archived.

We should talk about this feature's API and use cases. I have a few questions:

1. What is the use case for it? 
2. If it truly is a generalized feature that we should make available to the community, we should be careful with the API we want to expose. I don't think adding a one-off panel is a clear win.

jtpio:
Another thing that was discussed before was that the debugger sidebar might get a little bit busy if we start adding new panels to it (unless they are made collapsible - https://github.com/jupyterlab/debugger/issues/325)

mnowacki-b:
@afshin @jtpio 
Thank you for your response!

> We should not add any more features to the JupyteLab 2.x @jupyterlab/debugger 0.3.x line and we should add them directly to JupyterLab core

Understood, once we agree on the changes I can make the PR against the JupyterLab core. However, we would like to release this change soon, maybe even before JupyteraLab 3.x is released, so we are hope that we will be able to port it back.

> 1. What is the use case for it?

In our case we want to add an info panel which would contain information that will help with discoverability of the feature. We want to provide a short instruction on how to go about and enable the debugger (I don't have the precise wording yet).

Aside from that, I think it would be valuable to allow customization of the debugger sidebar where extensions could be made to either override current panels or supplement the sidebar with additional panels that could contain extra debugging info/tooling 

> 2. If it truly is a generalized feature that we should make available to the community, we should be careful with the API we want to expose. I don't think adding a one-off panel is a clear win.

Yes, after I entered this PR I thought that maybe it would have been better to have a provider for a sidebar panel factory. Do you think that would be a better route?

>Another thing that was discussed before was that the debugger sidebar might get a little bit busy if we start adding new panels to it (unless they are made collapsible - #325)

Yes we were thinking about making the custom panel collapsible because once the user is aware they would most likely want to reclaim the real estate. But having all panels be collapsible would be a great improvement. Also, it would be nice to have ability to rearrange them :).

jtpio:
> In our case we want to add an info panel which would contain information that will help with discoverability of the feature. We want to provide a short instruction on how to go about and enable the debugger (I don't have the precise wording yet).

Maybe the `Help` menu could be an appropriate spot to place such information, especially if it would be used only once or occasionally?

Another thing that comes to mind UX-wise to make it easier to discover features would be something like the `jupyterlab-tour` extension, that could for example highlight the debugger sidebar: https://github.com/fcollonval/jupyterlab-tour

> Yes we were thinking about making the custom panel collapsible because once the user is aware they would most likely want to reclaim the real estate. But having all panels be collapsible would be a great improvement. Also, it would be nice to have ability to rearrange them :).

Agree, being able to re-arrange and collapse the panels would definitely be useful :+1: 

Or being able to hide / show them with a right-click context menu, and persist the settings:

image

mnowacki-b:
> Maybe the Help menu could be an appropriate spot to place such information, especially if it would be used only once or occasionally?

In our case `Help` Manu may not be ideal. We would like to have an ability to alter the sidebar and potentially add extra panels whether it be an informational panel or something else. 
I think having flexibility to add custom panels would be useful as it will allow anyone to customize the Debugger sidebar to their needs.

Thoughts?

jtpio:
Thanks @mnowacki-b for taking the time to explain the use case.

> I think having flexibility to add custom panels would be useful as it will allow anyone to customize the Debugger sidebar to their needs.

This would be useful :+1: 

The `main` plugin could for example return a handle on the `Debugger.Sidebar`, and the sidebar would then expose an interface to add new panels.

But to echo @afshin's comment above we should indeed be careful to the API that will be exposed.

Otherwise as an alternative to the Help menu, there could also be a plugin that adds a new panel to the JupyterLab `right` area below the debugger sidebar (so it is located close to the debugger visually), for example with an "Info" icon (:information_source:):

image

mnowacki-b:
> The main plugin could for example return a handle on the Debugger.Sidebar, and the sidebar would then expose an interface to add new panels.

That sounds like a good idea. 
We would probably want something similar to the toolbar where we can either [addPanel](https://github.com/jupyterlab/jupyterlab/blob/4e7c446dccae5e82b1eba9af3eb1705917bc91a5/packages/apputils/src/toolbar.tsx#L219) (panel will be inserted at the end) or [insertPanel](https://github.com/jupyterlab/jupyterlab/blob/4e7c446dccae5e82b1eba9af3eb1705917bc91a5/packages/apputils/src/toolbar.tsx#L240) (panel will be inserted at a specific index).

@afshin @jtpio 
Any thoughts about that?

> Otherwise as an alternative to the Help menu, there could also be a plugin that adds a new panel to the JupyterLab right area below the debugger sidebar (so it is located close to the debugger visually), for example with an "Info" icon (ℹ️):

That is a good idea, but I am bit hesitant as the information are separate and the user will have to dig in to find the info. Where as when they open up the sidebar the information is right there.

jtpio:
> Any thoughts about that?

Sound like this would be more flexible indeed :+1: 

However since it's an API change, we'll want to be careful on how to get it in and released.

Ideally this would be added to JupyterLab directly, so we can go ahead with [archiving this repository](https://github.com/jupyterlab/debugger/issues/536) and keep discussions and PRs in one place.

mnowacki-b:
@jtpio Thanks!

I will try to make the code changes and create the PR against the JupyterLab (We may still need to merge those changes back into this repo depending on the JLab3.0 deployment timeline)

jtpio:
@mnowacki-b would you like to port the changes from https://github.com/jupyterlab/jupyterlab/pull/9007 to this repo, so we can cut a new release for JupyterLab 2.x?

mnowacki-b:
@jtpio 
Yes, will start working on it now.
Thank you!

mnowacki-b:
@afshin @jtpio 
I have ported over the changes form https://github.com/jupyterlab/jupyterlab/pull/9007 .
Since the repo is structured slightly differently I had to make some minor adjustments.
- added an export statement 
   https://github.com/jupyterlab/debugger/pull/533/files#diff-f41e9d04a45c83f3b6f6e630f10117feR60
- since string localization missing from this repo I made sure not to pull that change in.
   
Please have a look
Thank you!

jtpio (approved):
Thanks!

mnowacki-b:
Thank you!

PR #542, opened 2020-09-11
Do not minimize on Binder

jtpio:
Looks like recent versions of JupyterLab 2.x fail to build on Binder due to high memory consumption.

Turning off minimization should help make the build succeed.

https://mybinder.org/v2/gh/jupyterlab/debugger/binder-minimize-false?urlpath=/lab/tree/examples/index.ipynb


jtpio:
The Binder link posted above now builds correctly.

PR #549, opened 2020-10-08
Switch to a different murmurhash implementation to handle Unicode characters

### LINKED ISSUE #538: Gui error when using unicode symbols and setting a breakpoint
Hi!

i just found a minor bug within the the debugger's GUI, which occurs if you have a unicode symbol within the code of a cell. 
e.g.,: `print("€")`. 
Whenever this happens, the GUI is not able to display the breakpoint and the highlighting within a cell, however it is still registered in the window to the right and the debugger works properly. The breakpoints for the other cells still display as usual.

this is how it looks like (Line 52):
image

This is how I'm expecting it to behave (Line 52):
image

jtpio:
Fixes #538.

This change switches to using the murmurhash2 implementation from: https://github.com/jtpio/murmurhash2

### Before

unicode-before

### After

unicode-after

### The issue

Here is a report after digging into this and looking for different solutions. Posting a bit of information here so it's more convenient to refer to it in the future.

In https://github.com/jupyterlab/debugger/issues/538, we noticed that having non-ascii characters such as `€` in a code cell could be reproduced on the stable Binder (using `0.3.3`).

It was because the MurmurHash2 implementation in JS and the one used in `xeus-python` would not give the same result for the same input and seed.

`@jupyterlab/debugger` was using this implementation: https://github.com/mikolalysenko/murmurhash-js/blob/f19136e9f9c17f8cddc216ca3d44ec7c5c502f60/murmurhash2_gc.js#L14-L50, which corresponds to the version published on npm: https://www.npmjs.com/package/murmurhash-js 

`xeus-python` is using this implementation: https://github.com/xtensor-stack/xtl/blob/7d41f768787ee6405e3f1b1056e04f6fbd43f8cd/include/xtl/xhash.hpp#L47
Which is based on the original in C++: https://github.com/aappleby/smhasher/blob/master/src/MurmurHash2.cpp

Using this code snippet:

```python
print("€")
```

With `3339675911` as the seed.

Would give:

- `murmurhash-js`: `3511777296`
- `xtl`: `1079235191`

(more info in #538)

Summary:

image

From here two options:

**WebAssemly**

This sounds attractive because it would mean using the same code on both the backend and the frontend.
Reviving the initial work from https://github.com/xtensor-stack/xtl/pull/171 led to being able to use the `murmur2_x86` function in a node repl.  

However it started to feel a bit complicated when:

- instantiating WebAssembly on the web seems to be async only as of today. This would have to be taken into consideration when importing and loading the library, so the methods can be called when the instance is ready.
- packaging `.wasm` files for the web as part of an npm package is a bit complicated. At least as of today, but it might improve over time
- one option for packaging the wasm as js is to `base64` encode it, and create a new `Uint8Array` to be passed to `WebAssembly.initiate` in a `setup` hook / function or equivalent
- the bundle size from compiling `xtl/hash` to wasm with `emscripten` is in the order of ~100KB, which is quite a lot for just one function
- the `emscripten` / wasm tooling requires extra setup

**Fix the existing JS implementation**

The implementation from `murmurhash-js` states it is for ascii strings only: https://github.com/garycourt/murmurhash-js/blob/0197ce38bedac0e05f40b9d7152095d06db8292c/murmurhash2_gc.js#L9

Instead, we can convert the string to a `Uint8Array` via a `TextEncoder`, and perform the operations on the `Uint8Array` just like the original algorithm and the one used in `xeus-python`.
This is implemented in https://github.com/jtpio/murmurhash2

### Additional thoughts

Instead of having yet another `murmurhash2` package on npm, one way forward could be to move [the implementation](https://github.com/jtpio/murmurhash2/blob/0be3d062f024310dceba29d541469b8f67ef4bc9/src/index.ts) to [`@jupyterlab/coreutils`](https://github.com/jtpio/murmurhash2/blob/01a8ebc37164f7206abeb69a302c05ed5e233560/src/index.ts) instead.

The `0.3.x` releases of the debugger extension (this repo) could still depend on [`murmur2`](https://www.npmjs.com/package/murmurhash2).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ A kernel with support for debugging is required to be able to use the debugger.
 It is generally recommended to create a new `conda` environment to install the dependencies:
 
 ```bash
-conda create -n jupyterlab-debugger -c conda-forge xeus-python=0.8.0 notebook=6 jupyterlab=2 ptvsd nodejs
+conda create -n jupyterlab-debugger -c conda-forge xeus-python=0.8.6 notebook=6 jupyterlab=2 ptvsd nodejs
 conda activate jupyterlab-debugger
 ```
 
@@ -49,7 +49,7 @@ Enable the debugger, set breakpoints and step into the code:
 
 ```bash
 # Create a new conda environment
-conda create -n jupyterlab-debugger -c conda-forge nodejs xeus-python=0.8.0 ptvsd jupyterlab=2
+conda create -n jupyterlab-debugger -c conda-forge nodejs xeus-python=0.8.6 ptvsd jupyterlab=2
 
 # Activate the conda environment
 conda activate jupyterlab-debugger
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -7,4 +7,4 @@ dependencies:
 - nodejs
 - notebook=6
 - ptvsd
-- xeus-python=0.8.0
+- xeus-python=0.8.6
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,7 +12,8 @@ let local = {
   transform: {
     '\\.(ts|tsx)?$': 'ts-jest',
     '\\.svg$': 'jest-raw-loader'
-  }
+  },
+  testEnvironment: './test/custom-env'
 };
 
 Object.keys(local).forEach(option => {
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "@lumino/datagrid": "^0.6.0",
     "@lumino/disposable": "^1.2.0",
     "@lumino/widgets": "^1.8.0",
-    "murmurhash2": "^0.1.0",
     "vscode-debugprotocol": "^1.37.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "@lumino/datagrid": "^0.6.0",
     "@lumino/disposable": "^1.2.0",
     "@lumino/widgets": "^1.8.0",
-    "murmurhash-js": "^1.0.0",
+    "murmurhash2": "^0.1.0",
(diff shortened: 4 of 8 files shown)


afshin:
Hi @jtpio! How come the library is hosted on your account? Was it unavailable as a distribution by the original authors? Do you have a suggestion for how we should do this? We could bring the file into this repo, maybe?

jtpio:
> Was it unavailable as a distribution by the original authors? 

It's a modified version of the original to handle unicode. Also the one published to npm is not from the original authors but from a fork.

> We could bring the file into this repo, maybe

Yes I think so too. From the comment above:

> Instead of having yet another murmurhash2 package on npm, one way forward could be to move the implementation to @jupyterlab/coreutils instead.

afshin:
I missed that. That seems like a good idea to me. Or even including it as a local file we do not export right in the debugger package for now.

jtpio:
> Or even including it as a local file we do not export right in the debugger package for now.

I think we can vendor it in `@jupyterlab/debugger@0.3.x` yes, if we then add it to `@jupyterlab/coreutils` in core for 3.0.

afshin:
Well basically if no one else uses it right now, I'd prefer to keep the API surface area of `@jupyterlab/coreutils` smaller rather than larger. What do you think?

jtpio:
Yeah at first it felt like it should be "out there", since there doesn't seem to be a variation of murmurhash2 that handles it on npm.

But I agree that we can lower the maintenance burden by not exposing it and keep it private to the debugger package.

jtpio:
ok, so tests are also failing in https://github.com/jupyterlab/debugger/pull/550

jtpio:
Tests fixed in #553

afshin (approved):
Awesome, thank you @jtpio!

PR #556, opened 2020-10-12
Fix handling of source breakpoints

### LINKED ISSUE #555: BreakpointCellComponent error with xeus-python 0.8.6
Testing locally with the latest from c8686e7af76929511548a7fdb00c5a6d9ad91c69, creating a new environment from scratch.

Grabbing the latest packages including `xeus-python=0.8.6` and `xeus=0.24.2`, breakpoints seem to be breaking when they are added and the removed:

xeus-python-0-8-6-breakpoint

This doesn't seem to be an issue with `xeus-python=0.8.3` and `xeus=0.24.1`.

jtpio:
Fixes #555

Follow-up to https://github.com/jupyterlab/debugger/pull/514 and https://github.com/jupyter-xeus/xeus-python/pull/314, so the debugger frontend can handle the debug info reply properly.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -573,7 +573,11 @@ export class DebuggerService implements IDebugger, IDisposable {
         const { breakpoints, source } = val;
         map.set(
           source,
-          breakpoints.map(point => ({ ...point, verified: true }))
+          breakpoints.map(point => ({
+            ...point,
+            source: { path: source },
+            verified: true
+          }))
         );
         return map;
       },


jtpio:
Marked as draft for now as this issue is not happening in core lab.

It looks like it is related to the strict null checks (enabled in core but not here).

jtpio:
Which means we should move forward with archiving this repo since the code bases have diverged.

jtpio:
Let's get it in so the patch can be part of 0.3.4.

PR #562, opened 2020-11-18
Set terminateDebuggee to false

JohanMabille:
Stopping the debugger should not stop the kernel.

### DIFF
--- a/test/service.spec.ts
+++ b/test/service.spec.ts
@@ -173,6 +173,7 @@ describe('DebuggerService', () => {
     ].join('\n');
 
     let breakpoints: IDebugger.IBreakpoint[];
+    let sourceBreakpoints: IDebugger.IBreakpoint[];
     let sourceId: string;
 
     beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


jtpio:
Thanks!

This should fix the following behavior of the kernel restarting when the debugger is stopped:

xeus-python-restart

jtpio:
Looks like the tests are timing out, similar to https://github.com/jupyterlab/jupyterlab/pull/9362#issuecomment-729917352

test/service.spec.ts
```diff
@@ -190,6 +191,14 @@ describe('DebuggerService', () => {
           }
         };
       });
+      // create the list of source breakpoints expected from a restore
+      sourceBreakpoints = breakpoints.map(breakpoint => {
+        return {
+          line: breakpoint.line,
+          verified: breakpoint.verified,
+          source: breakpoint.source
```
jtpio, line 199: Ah, looks like this change is needed in this code base, because of this: https://github.com/jupyterlab/debugger/pull/556

And the fact they have diverged quite a bit because of the strict null checks in core but not here...

jtpio:
Let's get it in once the tests pass so we can cut a new patch release.

jtpio:
Thanks!

PR #566, opened 2020-11-26
Clear models and signals upon continue response

JohanMabille:
See https://github.com/jupyterlab/jupyterlab/pull/9402 for the explanation.


jtpio:
Thanks!

